Issue #17, opened 2024-10-05
Download Circuit JSON should work


seveibar:
Circuit JSON is already computed in the app so this should be easy to make "downloadable"

![Image](https://github.com/user-attachments/assets/d8d244c1-1b49-42fc-83f9-be5300240bef)

Abse2001:
commit #49

Issue #16, opened 2024-10-05
Download buttons should work


seveibar:
- [x] #17
- [x] #18
- [ ] Download kicad zip should work
- [ ] Download kicad_mod footprint should work

imrishabh18:
@seveibar By download you mean that a `.tsx` file should be downloaded and similarly a `.json` for circuit?

seveibar:
@imrishabh18 thinking about this more, I don't think we should support that as a feature because snippets will eventually be able to have multiple files and people can literally just ctrl-c/cmd-c in the short term. I removed that checklist item

imrishabh18:
Yeah, I thought by download you mean't to download the file. Which is not required as you can copy paste

Abse2001:
@seveibar  Im going to fix this issue to be clear no .tsx file because people can just copy and paste right ?

seveibar:
Yep!

PR #26, opened 2024-10-06
Feat: Download JSON button added

DhairyaMajmudar:
-> Download JSON button is now functional 
-> cloese #17
-> related to #16

cc: @seveibar


seveibar:
can we remove this button? as per the discussion here: https://github.com/tscircuit/snippets/issues/16#issuecomment-2395273258

You could easily adapt this PR to make the Download Circuit JSON button though!

src/components/DownloadButtonAndMenu.tsx
```diff
@@ -7,8 +7,18 @@ import {
...
 import { Download, ChevronDown } from "lucide-react"
+import { DownloadTSX } from "@/lib/utils/download-fn/download-tsx"
 
 export function DownloadButtonAndMenu({ className }: { className?: string }) {
+  const createBlobURL = (content: string) => {
+    const blob = new Blob([content], { type: "text/plain" })
+    return URL.createObjectURL(blob)
+  }
+
+  const { TSXContent, TSXFileName } = DownloadTSX()
```
seveibar, line 18: hooks should always have a `use*` prefix and be `camelCased`, likewise for variables returned

DhairyaMajmudar:
> can we remove this button? as per the discussion here: [#16 (comment)](https://github.com/tscircuit/snippets/issues/16#issuecomment-2395273258)
> 
> You could easily adapt this PR to make the Download Circuit JSON button though!

I got you let me modify this PR for downloading JSON data

src/components/DownloadButtonAndMenu.tsx
```diff
@@ -7,8 +6,18 @@ import {
...
 import { Download, ChevronDown } from "lucide-react"
+import { useDownloadJSON } from "@/hooks/use-download-json"
 
 export function DownloadButtonAndMenu({ className }: { className?: string }) {
+  const createBlobURL = (content: string) => {
+    const blob = new Blob([content], { type: "text/plain" })
+    return URL.createObjectURL(blob)
+  }
+
+  const { jsonContent, jsonFileName } = useDownloadJSON()
```
seveibar, line 17: Bad hook name, this hook should be named to indicate its used to download the current circuit json

useDownloadCurrentSnippetCircuitJson()

src/components/DownloadButtonAndMenu.tsx
```diff
@@ -7,8 +6,18 @@ import {
 } from "@/components/ui/dropdown-menu"
 import { Button } from "@/components/ui/button"
 import { Download, ChevronDown } from "lucide-react"
+import { useDownloadJSON } from "@/hooks/use-download-json"
 
 export function DownloadButtonAndMenu({ className }: { className?: string }) {
+  const createBlobURL = (content: string) => {
```
seveibar, line 12: Generic methods like this shouldnt be put in the inner scope of a function, move to utils directory

src/hooks/use-download-json.ts
```diff
@@ -0,0 +1,25 @@
...
+  const snippetId = useCurrentSnippetId()
+  const { data: snippet, isLoading } = useSnippet(snippetId)
+
+  const defaultCode = useMemo(() => {
+    return decodeUrlHashToText(window.location.toString()) ?? snippet?.code
+  }, [])
+
+  const [code, setCode] = useState(defaultCode ?? "")
+
+  const { message, circuitJson } = useRunTsx(code, snippet?.snippet_type)
```
seveibar, line 17: Very expensive recompute

seveibar (changes requested):
When the download button is clicked, compute the url and direct the user to download. Recomputing and rerunning typescript is too expensive

you can probably pass in the circuit json from the parent as a prop

src/hooks/use-download-json.ts
DhairyaMajmudar: I've installed the pacakage `file-saver` which will install the files

seveibar: the issue here is your running the tsx every time the page renders just to display a download button

seveibar: i think the easiest solution is to have the `circuitJson` passed into the `DownloadButtonAndMenu` and remove this hook

src/components/DownloadButtonAndMenu.tsx
```diff
@@ -7,8 +6,18 @@ import {
...
+import { createBlobURL } from "@/lib/createBlobURL"
+import create from "fake-snippets-api/routes/api/snippets/create"
 
 export function DownloadButtonAndMenu({ className }: { className?: string }) {
+  const { jsonContent, jsonFileName } = useDownloadCurrentSnippetCircuitJson()
+
+  const handleDownloadClick = () => {
+    const blob = createBlobURL(jsonContent)
+    saveAs(blob, jsonFileName)
+  }
```
seveibar, line 20:
```suggestion
export function DownloadButtonAndMenu({ className, circuitJson }: { className?: string, circuitJson?: AnyCircuitElement[] }) {

  const handleDownloadClick = () => {
    const blob = createBlobURL(JSON.stringify(circuitJson))
    saveAs(blob, jsonFileName)
  }
```

seveibar (commented):
see comments, we don't want to run the tsx (expensive operation) every time the download button renders

src/hooks/use-download-json.ts
DhairyaMajmudar: Cool! Thanks for the help let's do it this way.

seveibar:
This was completed in https://github.com/tscircuit/snippets/pull/49

PR #46, opened 2024-10-09
Initial ordering API and UI

andrii-balitskyi:
- **Add db resource and methods related to order and order_file resources**
- **Add /orders endpoints**
- **Add /order_files get and upload endpoints**
- **Add /my-orders page**
- **Add My order item to profile picture dropdown**
- **Add /orders/list endpoint**


src/components/Header.tsx
```diff
@@ -65,6 +65,9 @@ export default function Header() {
               <li>
                 <HeaderButton href="/ai">AI</HeaderButton>
               </li>
+              <li>
+                <HeaderButton href="/my-orders">My Orders</HeaderButton>
+              </li>
```
seveibar, line 70:
```suggestion
```

seveibar (approved):
looks good, note some minor changes where i removed it from the header, we don't want this feature to be super visible since snippets is publicly deployed and most of it works. It's still accessible by clicking the user dropdown menu

seveibar:
needs format

PR #49, opened 2024-10-10
Implemented a download function for the circuitJson download button.

Abse2001:
Issue #17


src/components/EditorNav.tsx
```diff
@@ -51,6 +52,7 @@ export default function EditorNav({
   onSave: () => void
 }) {
   const [, navigate] = useLocation()
+  const { circuitJson } = useRunTsx(code, snippet?.snippet_type)
```
seveibar, line 55: can you avoid using `useRunTsx`? It's a very expensive hook and the circuitJson should be computed elsewhere

seveibar (changes requested):
see comment, useRunTsx is pretty expensive, i would also recommend the `file-saver` module from @DhairyaMajmudar PR, you could also implement a different download format to avoid conflicts

src/components/EditorNav.tsx
```diff
@@ -111,7 +114,11 @@ export default function EditorNav({
           <Sparkles className="mr-1 h-3 w-3" />
           Edit with AI
         </Button>
-        <DownloadButtonAndMenu className="hidden md:flex" />
+        <DownloadButtonAndMenu
+          fileName={snippet.unscoped_name}
```
seveibar, line 118: important concept: "variable transarency" means you keep the same variable name as it is passed around. Don't rename variables if you don't have to. When I read this code, i was surprised that fileName was the same as the unscoped name, i expected it to have an extension. Naming is hard, renaming makes you have to do it more often

```suggestion
          snippetUnscopedName={snippet.unscoped_name}
```

src/lib/download-fns/download-circuit-json-fn.ts
```diff
@@ -0,0 +1,11 @@
+import { saveAs } from "file-saver"
+import { createBlobURL } from "./createBlobURL"
+export const downloadCircuitJson = (content: any, fileName: string) => {
+  try {
+    const circuitJson = JSON.stringify(content, null, 2)
+    const blob = new Blob([circuitJson], { type: "application/json" })
+    saveAs(blob, fileName)
+  } catch (error) {
+    throw error
+  }
```
seveibar, line 10: anti-pattern: useless try catch (it doesn't do anything here)

```suggestion
  const circuitJson = JSON.stringify(content, null, 2)
  const blob = new Blob([circuitJson], { type: "application/json" })
  saveAs(blob, fileName)
```

src/pages/view-snippet.tsx
```diff
@@ -45,7 +44,6 @@ export const ViewSnippetPage = () => {
                 <Share className="mr-1 h-3 w-3" />
                 Copy URL
               </Button>
-              <DownloadButtonAndMenu />
```
seveibar, line 48: why remove from this page? can't we give it the props?

seveibar (changes requested):
see comments

DhairyaMajmudar:
It seems like I've already implemented this feature here https://github.com/tscircuit/snippets/pull/26

seveibar:
@DhairyaMajmudar for small features like this it's just whoever gets it right first, I don't do assignment unless it's a big feature. In open-source a project can be completely stalled when people don't fix PRs quickly etc. so we won't operate where people can work on things first

The best way to get issues assigned exclusively to you is to get familiar with the codebase and work on hard issues that people don't generally want, or find bugs and create your own issues.

seveibar (approved):

PR #56, opened 2024-10-13
allow anon mutations for logged in users, generally don't create snippet unless save button is pressed, various fixes, refactor for reuse, playwright snapshot testing, loading/404 pages

seveibar:


seveibar:
update snapshots

PR #82, opened 2024-10-17
Allow editing shipping information in profile settings

andrii-balitskyi:
Closes https://github.com/tscircuit/snippets/issues/72

- **Add /accounts/get and /update to fetch/update shipping info**
- **Update db schema adding shipping info**
- **Add ShippingInformationForm component and use it on the setting page**

https://github.com/user-attachments/assets/4fae296f-c70e-4336-a6e4-3ab7684ae628


seveibar (approved):
Nice work. We’ll need to make country and state a dropdown, can be a future issue

PR #86, opened 2024-10-18
Update page title to snippet name.

### LINKED ISSUE #81: Change page title when viewing a snippet or editing a snippet to have snippet name
`tscircuit - <owner>/<package>`

kom-senapati:
Closes #81

Shows name of snippet and owner name in title as format `tscircuit - <owner>/<package>`


seveibar (changes requested):
Can you use react helmet so that other metadata can be added and the pattern is a bit more reusable. Looks good though, just probably a good opportunity to set up a pattern

seveibar (approved):
nice work!

PR #95, opened 2024-10-19
fixed:quickstart page responsiveness

theajmalrazaq:
just fixed the quickstart page responsivess it is subpart of issue #25:mobile view is broken
Screenshot_select-area_20241019124711
Screenshot_select-area_20241019124722

PR #97, opened 2024-10-20
Added:Playwright test for md,lg and sm

theajmalrazaq:
So, what I did is I wrote Playwright tests for the AI Page and Quickstart Page, added viewports to the existing tests, fixed them, and updated the snapshots."


seveibar (approved):
Nice work. I think there is an issue with lg where the grid is supposed to be four columns but is 3, not sure, but i like all the snapshots

theajmalrazaq:
@seveibar There used to be four, but I reduced the number to three to enhance the user interface.

seveibar:
@theajmalrazaq its worse on lg. can you switch it to 4 on lg?

I didnt see text overflowing

playwright-tests/ai-page.spec.ts
```diff
@@ -1,10 +1,19 @@
...
+    await page.setViewportSize(viewport)
+    await page.goto("http://127.0.0.1:5177/ai")
+
+    // it is only working for desktop beacuse fake testuser login is not appearing for mobile
+
+    // await page.goto("http://127.0.0.1:5177");
+    // await page.click("text='Fake testuser Login'");
+    // await page.waitForSelector(".login-avatar");
+    // await page.click(".header-button[href='/ai']");
+    // await page.waitForSelector("text='Submit a prompt to get started!'");
```
seveibar, line 15: Issue: Get ai page working on mobile

theajmalrazaq: @seveibar my PR is still under review #95 i'll change it to 4 cols after that i'll inform you.

theajmalrazaq: > Issue: Get ai page working on mobile

It taking snapshot perfectly but the problem is that fake tester login button only shows on lg devices in navbar so my can't take snapshot in mobile devices.

PR #121, opened 2024-10-23
Fixes Command palette search

### LINKED ISSUE #119: CmdK Menu Doesn't Search Snippets Properly (no results when searching)
I've reproduced this several times, the react-query setup is likely wrong, it needs to keep searching with a queryKey that is unique for the search string

Image

mrudulpatil18:
Fixes #119 

Changes:
- Rewrote a bunch of logic to fix the search filtering issue
- Now uses Fuzzy search to filter possible commands in the CMD-K menu
- This fixes the searching results not shown issue

cc: @seveibar


mrudulpatil18:
NOTE: still requires a bit of work with a weird bud and some refactoring, will do when i get some free time

seveibar (approved):

seveibar (approved):
Sounds good! Yea i prefer the cmdk library so i support the change

seveibar:
@mrudulpatil18 can we merge to capture these changes?

mrudulpatil18:
> @mrudulpatil18 can we merge to capture these changes?

Yup, it's working as of now

mrudulpatil18:
@seveibar 
added absolute position for the cmd-k dialog

PR #134, opened 2024-10-29
feat: manual-edits.json is now editable

imrishabh18:
Screenshot 2024-10-29 at 5 07 07 PM


src/components/CodeEditor.tsx
```diff
@@ -238,42 +276,137 @@ ${code}
...
+  }, [files, currentFile])
 
-  const codeImports = getImportsFromCode(code)
+  const codeImports = getImportsFromCode(files["index.tsx"])
 
   useEffect(() => {
-    if (ataRef.current) {
-      ataRef.current(code)
+    if (ataRef.current && currentFile === "index.tsx") {
+      ataRef.current(`${DEFAULT_IMPORTS}${files["index.tsx"]}`)
```
seveibar, line 305: I think we should keep the "code" alias to simplify the diff imo

seveibar (approved):
Note before merging: there is an important change that will break production with the apiUrl, make sure to revert that line!


Look good, i think more code is now in CodeEditor when our goal should be to reduce the size of CodeEditor since its so complex. Could we make the CodeEditorHeader a separate component to hopefully start to break down that massive component?

seveibar:
Another random note, this code can be a bit sensitive to changes, so definitely test stuff like editing code and making sure new typescript imports are recognized and dynamically imported etc. also that the cursor is stable when editing.

We could potentially incorporate that into a playwright test later

Issue #109, opened 2024-10-21
Warn user if they leave the `/editor` page with unsaved changes


seveibar:

KhushiSharma0313:
Can I work on it?

PR #135, opened 2024-10-29
feature: Warn user if they leave the /editor page with unsaved changes #109

Satvik1769:
This pr solves issue #109 
Using beforeUnload and popStatus event I solved the issue and tested it in various ways manually.
Please merge this pr as soon as possible.

Regards,
Satvik


seveibar:
@Satvik1769 can you fix the conflict, then we can run workflows

seveibar:
It might be because of another PR being merged but you still have conflicts.

You don't need to remove the bun.lockb (that's actually bad and will cause issues), just run `bun i` and any conflicts in the existing bun.lockb will automatically be fixed by bun

Issue #129, opened 2024-10-28
When using "Copy URL" for a snippet, store the type of snippet in the url i.e. "board" or "module"


seveibar:
This will prevent the error when you copy the url for a module it is incorrectly loaded as a board when you visit the url.

Alternatively if the url parameter isn't specified, we should set the type to "module" if we do not see "<board" in the code

PR #137, opened 2024-10-29
adds snippet type to the url for copy url

mrudulpatil18:
Issue: #129

- Updated encodeTextToUrlHash to add snippet type to the URL.
- If not explicitly specified : Defaults to board if <board is detected in code; otherwise package.

### DIFF
--- a/src/lib/encodeTextToUrlHash.ts
+++ b/src/lib/encodeTextToUrlHash.ts
@@ -1,7 +1,7 @@
 import { gzipSync, strToU8 } from "fflate"
 import { bytesToBase64 } from "./bytesToBase64"
 
-export function encodeTextToUrlHash(text: string): string {
+export function encodeTextToUrlHash(text: string, type?: string): string {
   // Compress the input string
   const compressedData = gzipSync(strToU8(text))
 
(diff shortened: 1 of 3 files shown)


src/lib/encodeTextToUrlHash.ts
```diff
@@ -1,13 +1,14 @@
 import { gzipSync, strToU8 } from "fflate"
 import { bytesToBase64 } from "./bytesToBase64"
 
-export function encodeTextToUrlHash(text: string): string {
+export function encodeTextToUrlHash(text: string, type?: string): string {
```
seveibar, line 4:
```suggestion
export function encodeTextToUrlHash(text: string, snippet_type?: string): string {
```

src/lib/encodeTextToUrlHash.ts
```diff
@@ -1,13 +1,14 @@
...
   // Compress the input string
   const compressedData = gzipSync(strToU8(text))
 
   // Convert to base64
   const base64Data = bytesToBase64(compressedData)
 
   // Construct the URL
-  return `${window.location.origin}/editor#data:application/gzip;base64,${base64Data}`
+  const typeParam = type ? `&type=${type}` : ""
```
seveibar, line 12:
```suggestion
  const typeParam = snippet_type ? `&snippet_type=${type}` : ""
```

seveibar (approved):
always better to keep prefixes with `type`, `data` or other vars that are often ambiguous in different contexts. e.g. we could support `order_type`, `template_type` or other url params in the future with that kind of name

src/lib/encodeTextToUrlHash.ts
mrudulpatil18: modified in recent commit

src/lib/encodeTextToUrlHash.ts
mrudulpatil18: modified in recent commit

PR #139, opened 2024-10-30
feature:  Warn user if they leave the /editor page with unsaved changes #109

Satvik1769:
This pr solves issue https://github.com/tscircuit/snippets/issues/109
Using beforeUnload and popStatus event I solved the issue and tested it in various ways manually.
I believe conflicts will not occur now
Please merge this pr as soon as possible.

Regards,
Satvik


seveibar:
@Satvik1769 failing format check, you can fix with `bun format`

src/components/CodeEditor.tsx
```diff
@@ -221,7 +227,7 @@ export const CodeEditor = ({
             hoverTooltip((view, pos, side) => {
               const { from, to, text } = view.state.doc.lineAt(pos)
               const line = text.slice(from, to)
-              const match = line.match(/@tsci\/[\w\-.]+/)
+              const match = line.match(/@tsci\/[\w.]+/)
```
seveibar, line 230: you reverted some fixes here. Make sure to only commit changes that are relevent for your fix

src/components/CodeEditor.tsx
```diff
@@ -331,6 +337,43 @@ export const CodeEditor = ({
...
+    if (hasUnsavedChanges) {
+      window.history.pushState(null, "", window.location.href)
+    }
+
+    // Cleanup event listeners on component unmount
+    return () => {
+      window.removeEventListener("beforeunload", handleBeforeUnload)
+      window.removeEventListener("popstate", handlePopState)
+    }
+  }, [hasUnsavedChanges])
```
seveibar, line 375: this component (the `CodeEditor`) is already really big, can you put this code in the parent component? Also can you make a dedicated hook e.g. `useWarnUserForUnsavedChanges(hasUnsavedChanges)`

Satvik1769:
Ok will fix the issues

src/components/CodeEditor.tsx
Satvik1769: i created  a hook for the functionality now we just have to call it. it is in hooks folder with name useWarnUserForUnsavedChanges. I also reverted this code to previous state i.e.
const match = line.match(/@tsci\/[\w\-.]+/)

I hope there is no more conflicts and hope it helps.

src/components/CodeEditor.tsx
```diff
@@ -323,6 +328,10 @@ export const CodeEditor = ({
     }
   }, [!isStreaming, currentFile, code !== ""])
 
+  useWarnUserForUnsavedChanges({
+    hasUnsavedChanges: hasUnsavedChanges ?? false,
+  })
```
seveibar, line 333: why does this have to be in the CodeEditor, can you move it to CodeAndPreview? This file is already very large and I don't see why this functionality has to be in here

src/hooks/useWarnUserForUnsavedChanges.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    const handlePopState = (event: PopStateEvent) => {
+      if (hasUnsavedChanges) {
+        const userConfirmed = window.confirm(
+          "Are you sure you want to go back? Unsaved changes may be lost.",
+        )
+        if (!userConfirmed) {
+          // Push the state again to stay on the same page
+          window.history.pushState(null, "", window.location.href)
```
seveibar, line 22: do we have to push state to stay on the same page? Can we just not do anything and block the action?

src/hooks/useWarnUserForUnsavedChanges.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+
+    // Attach event listeners
+    window.addEventListener("beforeunload", handleBeforeUnload)
+    window.addEventListener("popstate", handlePopState)
+
+    // Push the initial state only once, when component mounts
+    if (hasUnsavedChanges) {
+      window.history.pushState(null, "", window.location.href)
```
seveibar, line 33: why do we need to push state here?

src/components/CodeEditor.tsx
Satvik1769: Yes it can also be moved in CodeAndPreview and it will work fine.

src/hooks/useWarnUserForUnsavedChanges.ts
Satvik1769: state is necessary here this solution works by storing the current url in the state and if user clicks on cancel instead of sending it to previous page it send it to the current page

seveibar: What would happen if you removed the pushState?

seveibar: if i understand your code correctly, `preventDefault` stops the page from changing, so I'm not sure why you need to pushState

Satvik1769: It basically stores the current url (editor page) and only allows when cancelled. Is there a problem with storing the data in the state?

seveibar: Generally we only introduce state when absolutely necessary. If it's not necessary, we shouldn't do it.

I'm not convinced it's necessary. You only need pushState to change to a new page, and we're cancelling a navigation, so there should never be a need to pushState

src/hooks/useWarnUserForUnsavedChanges.ts
Satvik1769: we push the states two times:
1. At the starting of mounting if there is some unsaved changes
2. if person goes black but cancels then we repush the state to prevent moving back

src/hooks/useWarnUserForUnsavedChanges.ts
Satvik1769: the thing is preventDefault will work for case like reloading but in case of user going back it will not work

Satvik1769: we can try to do this with pop state functionality but it doesn't block the action rather using it we can push state to restore current page

Satvik1769: I believe to add this functionality we will need to have states involved. However, if you have better method in mind I can make it .

seveibar: let's remove the pushStates, if we only work for reloading that's OK for now.

Satvik1769: as you wish but i will just comment out that part just in case you want to update

seveibar: let's just remove it, i'd rather not have ghost code, people can come back and look at this PR.

Also make sure to revert the changes to files where things were accidentally changed. I'll leave some comments that indicate where that happened. There should be no changes to `CodeEditor` (IIUC), you can run `git checkout origin/main -- path/to/file` to revert a file

src/components/CodeEditor.tsx
```diff
@@ -128,9 +128,13 @@ export const CodeEditor = ({
             .replace(registryPrefixes[2], "")
           const packageName = fullPackageName.split("/")[0].replace(/\./, "/")
           const pathInPackage = fullPackageName.split("/").slice(1).join("/")
-          const jsdelivrPath = `${packageName}${pathInPackage ? `/${pathInPackage}` : ""}`
+          const jsdelivrPath = `${packageName}${
```
seveibar, line 131: revert unnecessary change here

src/components/CodeEditor.tsx
```diff
@@ -128,9 +128,13 @@ export const CodeEditor = ({
...
+            pathInPackage ? `/${pathInPackage}` : ""
+          }`
           return fetch(
-            `${apiUrl}/snippets/download?jsdelivr_resolve=${input.includes("/resolve/")}&jsdelivr_path=${encodeURIComponent(jsdelivrPath)}`,
+            `${apiUrl}/snippets/download?jsdelivr_resolve=${input.includes(
+              "/resolve/",
+            )}&jsdelivr_path=${encodeURIComponent(jsdelivrPath)}`,
```
seveibar, line 137: revert unnecessary change here

src/components/CodeEditor.tsx
```diff
@@ -263,7 +267,7 @@ export const CodeEditor = ({
                 for (let pos = from; pos < to; ) {
                   const line = view.state.doc.lineAt(pos)
                   const lineText = line.text
-                  const matches = lineText.matchAll(/@tsci\/[\w\-.]+/g)
+                  const matches = lineText.matchAll(/@tsci\/[\w.]+/g)
```
seveibar, line 270: this would introduce a bug, revert

src/hooks/useWarnUserForUnsavedChanges.ts
Satvik1769: ok will do

src/components/CodeEditor.tsx
Satvik1769: I am sorry for all this trouble i am thinking to shut this pr and create a cleaner pr that is easier to manage for you and ofcourse myself

seveibar: yes that's always ok!

seveibar: and no problem- my messages are short because i often type them from my phone, not because i'm angry 😁

PR #143, opened 2024-10-30
feat: Position update on movement in manual-edits

imrishabh18:
https://github.com/user-attachments/assets/9a122ee0-e8d0-4a8a-8cd4-154569e5bf62


src/components/CodeAndPreview.tsx
```diff
@@ -35,7 +35,7 @@ export function CodeAndPreview({ snippet }: Props) {
       templateFromUrl.code
     )
   }, [])
-  const [manualEditsText, setManualEditsText] = useState("")
+  const [manualEditsJson, setmanualEditsJson] = useState("")
```
seveibar, line 38:
```suggestion
  const [manualEditsJson, setManualEditsJson] = useState("")
```

src/components/CodeEditor.tsx
```diff
@@ -50,7 +52,15 @@ export const CodeEditor = ({
...
-    "manual-edits.json": "",
+    "manual-edits.json": JSON.stringify(
+      {
+        pcb_placements: [],
+        edit_events: [],
+        manual_trace_hints: [],
+      },
+      null,
+      2,
+    ),
```
seveibar, line 63: this eventually needs to be loaded from the snippet, but this is a great mergable start!

src/components/CodeEditor.tsx
```diff
@@ -65,6 +75,26 @@ export const CodeEditor = ({
     }
   }, [initialCode])
 
+  useEffect(() => {
+    if (manualEditsJson) {
+      setFiles((prev) => ({
+        ...prev,
+        "manual-edits.json": manualEditsJson,
+      }))
```
seveibar, line 83: technically this is a re-render anti-pattern, if `files` is completely derived you should use `useMemo` for it rather than the antipattern "`useEffect` with `setState`" which causes unnecessary re-renders and can complicate the rendering flow.

Can you say something like:

```tsx
const files = useMemo(() => ({
  "index.tsx": code,
  "manual-edits.json": manualEditsJson
}, [code, manualEditsJson])
```

`useMemo` can't cause re-renders, so it's much safer.

src/components/CodeEditor.tsx
```diff
@@ -276,7 +306,7 @@ export const CodeEditor = ({
     return () => {
       view.destroy()
     }
-  }, [code !== "", !isStreaming])
+  }, [!isStreaming, currentFile])
```
seveibar, line 309: i think this re-introduces a bug i fixed. We want to trigger the `useEffect` when the code goes from "" -> anything because that means that the code has loaded

```suggestion
  }, [!isStreaming, currentFile, code !== ""])
```

src/components/PreviewContent.tsx
```diff
@@ -26,6 +27,8 @@ export interface PreviewContentProps {
   isStreaming?: boolean
   onCodeChange?: (code: string) => void
   onDtsChange?: (dts: string) => void
+  manualEditsJson: string
+  onmanualEditsJsonChange: (newmanualEditsJson: string) => void
```
seveibar, line 31:
```suggestion
  onManualEditsJsonChange: (newManualEditsJson: string) => void
```

src/lib/utils/pcbManualEditEventHandler.ts
```diff
@@ -0,0 +1,148 @@
...
+  manual_trace_hints: ManualTraceHint[]
+}
+
+export const createInitialManualEditState = (): ManualEditState => ({
+  pcb_placements: [],
+  edit_events: [],
+  manual_trace_hints: [],
+})
+
+export const handlePcbEditEvents = (
```
seveibar, line 29: `handle` implies effects, `apply` is a better prefix here.

```suggestion
export const applyEditEvents = (
```

seveibar (approved):
see comments, react hell is `useEffect` bugs. Recommendation: Insert playwright tests (aider is quite good at writing them) to make sure we don't have anything breaking for some of the use cases you're video'ing. Component movement might be difficult to capture because it acts on a canvas, so i don't think that's possible for now unless we do absolute coordinates against the canvas "drag from x,y to x,y relative to the top left coordinate of the canvas"

It would also be a good idea to start counting re-renders inside the codeeditor but I can't remember the pattern for doing that.

src/lib/utils/pcbManualEditEventHandler.ts
seveibar: `apply` can imply mutations, so it's not perfect fwiw, but `handle` is broader and can imply "global effects" in a react context, e.g. binding to window

Issue #130, opened 2024-10-28
Profile page text update header


imrishabh18:
The Header is incorrect, it should be something like `My Profile` or `Profile`. And, the link redirected by the button `View Github` is redirecting to this url https://github.com/profile  and not the profile of the user.

![Image](https://github.com/user-attachments/assets/26be3a66-6b4e-4a8b-9a80-2f1210dbb365)

SufyaanKhateeb:
I can work on this.

imrishabh18:
We are not assigning issues, until you have decent contribution in this repository. But feel free to work on this and open a PR

SufyaanKhateeb:
This issue seems to occur only if you navigate to the profile page using the footer link. This footer "My Profile" link is visible even if the user is not logged in, so what is the expected behaviour on clicking the link in that case or should I just hide the button in that case?

The footer link redirects you to the /profile if we just fix the redirect to go to the /{username} this problem will be fixed and also in this case the header text will be {username}'s profile, do you want it to be "My profile" or "Profile" if the user is in their profile page?

imrishabh18:
The profile link should not be visible in the footer when the user is not logged in.

cc: @seveibar

SufyaanKhateeb:
@imrishabh18 or @seveibar Can you please add the hacktoberfest-accepted label to this issue.

imrishabh18:
@SufyaanKhateeb Not needed, this repository has the `hacktoberfest` label. When you PR gets merged it will be counted towards the Hacktoberfest

SufyaanKhateeb:
It has been merged BTW. You can close this issue if it is OK.

PR #144, opened 2024-10-30
fix(ui): fix ui for profile page header and footer "My profile" link

SufyaanKhateeb:
Fix for #130:
Fix the profile page header to say "My Profile" when the current user is viewing their profile and fix the "My Profile" link in the footer to show up only when logged in.


seveibar (approved):
i like it

PR #161, opened 2024-11-01
feat: Manual edit props getting passed to index.tsx

imrishabh18:
https://github.com/user-attachments/assets/3220c355-2e12-40fe-8d76-7cb4fce6a911


src/components/CodeAndPreview.tsx
```diff
@@ -57,7 +61,7 @@ export function CodeAndPreview({ snippet }: Props) {
     triggerRunTsx,
     tsxRunTriggerCount,
   } = useRunTsx({
-    code,
+    code: importManualEditsCheck(code, manualEditsJson),
     type: snippetType,
   })
```
seveibar, line 66:
```tsx
const importsMap = useMemo(() => ({
  "./manual-edits.json": JSON.stringify(manualEditsJson)
}, [manualEditsJson])
const {
    message,
    circuitJson,
    compiledJs,
    triggerRunTsx,
    tsxRunTriggerCount,
  } = useRunTsx({
    code,
    importsMap,
    type: snippetType,
  })
```

seveibar: variable transparency violation, `code` should not change meaning

seveibar: actually this isn't right either, one sec

src/components/CodeAndPreview.tsx
```diff
@@ -35,7 +36,9 @@ export function CodeAndPreview({ snippet }: Props) {
       templateFromUrl.code
     )
   }, [])
-  const [manualEditsJson, setManualEditsJson] = useState("")
+  const [manualEditsJson, setManualEditsJson] = useState(
```
seveibar, line 39: If it's a string, we should make sure the variable name makes it clear it's a string.

```suggestion
  const [manualEditsFileContent, setManualEditsFileContent] = useState(
```

src/hooks/use-run-tsx/index.tsx
```diff
@@ -21,10 +19,12 @@ type RunTsxResult = {
 
 export const useRunTsx = ({
   code,
+  userImports,
   type,
   isStreaming = false,
 }: {
   code?: string
+  userImports?: Record<string, string>
```
seveibar, line 27: If i understand your change correctly, userImports should map to an object now right?

```suggestion
  userImports?: Record<string, object>
```

src/components/CodeAndPreview.tsx
```diff
@@ -50,6 +53,13 @@ export function CodeAndPreview({ snippet }: Props) {
 
   const { toast } = useToast()
 
+  const importsMap = useMemo(
```
seveibar, line 56: variable transparency

```suggestion
  const userImports = useMemo(
```

src/hooks/use-run-tsx/index.tsx
```diff
@@ -64,21 +64,27 @@ export const useRunTsx = ({
...
+      const tscircuitRequire = (name: string) => {
         if (!preSuppliedImports[name]) {
           throw new Error(
             `Import "${name}" not found (imports available: ${Object.keys(preSuppliedImports).join(",")})`,
           )
         }
         return preSuppliedImports[name]
       }
-      ;(globalThis as any).__tscircuit_require = __tscircuit_require
+      ;(globalThis as any).__tscircuit_require = tscircuitRequire
```
seveibar, line 81: we should revert this, variable transparency (generally changing the name of a variable that represents the same thing should be avoided)

src/hooks/use-run-tsx/index.tsx
```diff
@@ -147,11 +153,11 @@ export const useRunTsx = ({
 
         const primaryKey = componentExportKeys[0]
 
-        const UserElm = (props: any) =>
+        const userElm = (props: any) =>
```
seveibar, line 156: we should revert this by React convention- components should be PascalCase

```suggestion
        const UserElm = (props: any) =>
```

src/hooks/use-run-tsx/index.tsx
imrishabh18: I think biome did this, will update

src/hooks/use-run-tsx/index.tsx
imrishabh18: I think these changes needs to be added to biome, else on save we will get this

seveibar (approved):
i have a feeling there's a bug when dragging based on the flashing in the screenshot. We need to make sure we're only using edit events that are "complete", otherwise we might be re-rendering the circuit every 10ms, the larger boards take 1-4s to render so starting hundreds of renders will cause the browser to crash or become unresponsive. I'm not 100% sure but i think we need to check that `editEvent.is_complete` before incorporating the edit event into the file. Just a hunch

src/hooks/use-run-tsx/index.tsx
seveibar: just `// @biome-ignore`, but idk how this file passed in the past if biome doesn't accept it?

imrishabh18:
I won't be merging this yet, even I have seen one bug in the pcbViewer edit events. Will look into it @seveibar

src/hooks/use-run-tsx/index.tsx
seveibar: no worries!

src/lib/utils/pcbManualEditEventHandler.ts
```diff
@@ -46,22 +46,22 @@ export const applyPcbEditEvents = (
...
         handledEventIds.add(event.edit_event_id)
       }
     })
 
     // Process new edit events
-    for (const editEvent of editEvents) {
+    for (const editEvent of changedEditEvents) {
       if (
-        !editEvent.edit_event_id ||
+        (editEvent.in_progress && !editEvent.edit_event_id) ||
```
imrishabh18, line 64: @seveibar this condition is checking for the events that are complete, and only then proceed. Thus reduces the re-renders.

seveibar: nice

seveibar:
merged as part of #162!!!

Issue #151, opened 2024-10-31
Long Code Snippets Cannot Be Viewed (Overflow issue)


seveibar:
Example: Not all the code is showing here: https://snippets.tscircuit.com/seveibar/esp32-s3-mini-1-n8

PR #164, opened 2024-11-02
Fix/overflow view page

imrishabh18:
Closes - #151 

https://github.com/user-attachments/assets/88644902-28a3-4406-a206-d6a79e219dfa


src/components/PreviewContent.tsx
```diff
@@ -80,7 +80,7 @@ export const PreviewContent = ({
         <Tabs
           value={activeTab}
           onValueChange={setActiveTab}
-          className="flex-grow flex flex-col"
+          className="flex-grow flex flex-col sticky top-0 h-screen"
```
seveibar, line 83: hmm is this needed? the container already has sticky

seveibar:
@imrishabh18 future pushes will deploy the preview

seveibar:
I want to play with this PR, for some reason the video doesn't load and I'm confused by the sticky and overflow parameters

src/components/PreviewContent.tsx
imrishabh18: Yes, that container sticky even when provided with the `h-screen` it doesn't make it stick. I think we can remove that div as a whole

https://github.com/user-attachments/assets/75a88b52-774e-46fd-92e9-41881bfdad8f

seveibar:
@imrishabh18 instead of having this overflow with scroll, can we make it expand to the full size of the page the same way it does on the editor page?


image

src/components/PreviewContent.tsx
seveibar: it might be my computer but the videos you're posting are 0s, is it just me?

seveibar: I still don't understand what this line fixes though, it's already in a sticky container. Maybe the video would help (let me know if it's on my end)

imrishabh18:
@seveibar You mean something like this?


https://github.com/user-attachments/assets/1f9e7376-28d1-43a0-9fcc-76df1c9324bd

seveibar:
> @seveibar You mean something like this?
> 
>  Screen.Recording.2024-11-03.at.3.16.52.AM.mov

Yep! I think that is more natural- it's what github does after all!

src/components/PreviewContent.tsx
imrishabh18: I think on my end, the video recorded was playing in my machine. But on the upload preview it doesn't

seveibar: Very strange. I use [Kap](https://getkap.co/) for recording fwiw

Issue #175, opened 2024-11-07
Fix schematic zooming


seveibar:
When you zoom in on the schematic view, it doesn't zoom into the correct location. This most likely requires fixing `use-mouse-matrix-transform` (do not replace it- I don't want to get a bunch of mouse math into this project unnecessarily, but a PR fixing that project is acceptable for the bounty)

In this video I'm scrolling up and down, it SHOULD zoom into the mouse location but instead the screen swings somewhat wildly.

https://github.com/user-attachments/assets/e4627984-ddb9-4f0c-9cc7-996c8d55e45b

RohittCodes:
Hey @seveibar, I would like to work on this. I see the issue is with the `CircuitToSvgWithMouseControl` component, but the actual `Schematic` component isn't stabilized and doesn't have the issue.

So, should I work with the `CircuitToSvgWithMouseControl` component?

seveibar:
The Schematic component is too hard to maintain because it uses a more complex rendering engine, it doesnt have support for rendering circuit json properly, so the fix should be with the CircuitSvgWithMouseControl component

PR #178, opened 2024-11-08
fix #175: zoom effect for schematic view

RohittCodes:
https://github.com/user-attachments/assets/c012a2bb-ac1c-4c19-ae47-6438f509b956

/claim #175

fixes: #175


RohittCodes:
Lemme know if it's okay, I forgot this point,
> I don't want to get a bunch of mouse math into this project unnecessarily

I can iterate over the solution and will try to add the same to `useMouseMatrixTransform` and respective functions if required. I've not replaced the `useMouseMatrixTransform`, but included a little math to it, which isn't that complex, so I guess it won't be a problem.

seveibar (changes requested):
well done! Yes ideally this math is included in the `useMouseMatrixTransform` library- otherwise it's difficult for others to maintain who are less familiar with matrix math

Also, to get the checks to pass just run `bun format`.

Thanks really awesome work!

RohittCodes:
okay sure!

RohittCodes:
@seveibar won't this be ideal if I work on the particular package, where the function is implemented? or is it good to be merged?

I'm talking about this repo: [use-mouse-matrix-transform](https://github.com/seveibar/use-mouse-matrix-transform)

seveibar:
Yea this is in part why I increased the bounty. First you have to reproduce the issue inside of `use-mouse-matrix-transform` (it has a way of creating examples if i remember correctly) then you have to fix the reproduction. The upside of doing this is nobody has to worry about this issue anymore because we use `use-mouse-matrix-transform` a lot

seveibar:
I just verified that the issue seems to be present in that repo's example so you may not need to make a reproduction!

image

Satvik1769:
Hi, I solved this issue in the use-mouse-matrix-transform by zooming the only required element to zoom instead of zooming every part of canvas. I hope this help please check out issue #3 of use-mouse-matrix-transform.

RohittCodes:
i guess the issue isn't with the `use-mouse-matrix-transform` I've updated it without any matrix-math logic this time, it updates the containerBounds, and handles the interactions based on position and the dimensions.
```
  const updateContainerBounds = useCallback(() => {
    if (!containerRef.current) return
    const rect = containerRef.current.getBoundingClientRect()
    containerBoundsRef.current = {
      width: rect.width,
      x: rect.left,
      y: rect.top,
    }
    setContainerWidth(rect.width)
  }, [])
```

the issue which the example in the referenced repo was facing because of the `relative` property and due to the modification in dimensions using `transform.d`.
here's more about it: 
https://github.com/seveibar/use-mouse-matrix-transform/pull/4#issuecomment-2466589666

seveibar:
@RohittCodes the code looks ok, but are all these changes necessary? Can you reduce it to the minimum set of things that fix the issue? If it's just adding `relative`, that's better than including unneeded changes

seveibar:
confirmed fix on preview, waiting to see if code can be reduced

RohittCodes:
yeah reduced it! it was just the `transformOrigin` property

RohittCodes:
@seveibar you can review it. it was just the css property which we were missing most probably.

Issue #79, opened 2024-10-16
Reduce size of search text


seveibar:
![Image](https://github.com/user-attachments/assets/c502c322-0466-422c-829b-e978f3e9348f)

arpitkuriyal:
I can work on it can you assign it to me?

RohittCodes:
@seveibar is the issue for search text or do you want the search results to be truncated or just the size to be reduced?

RohittCodes:
@seveibar do you want the size of the search results to be minimized or should i truncate it?

> Reduce size of search text

or is it just search text (inside the input)?

PR #181, opened 2024-11-10
Reduced size of search text

anas-sarkez:
/claim #79
#79

### DIFF
--- a/src/components/SearchComponent.tsx
+++ b/src/components/SearchComponent.tsx
@@ -57,7 +57,7 @@ const SearchComponent: React.FC<SearchComponentProps> = ({
       <Input
         type="search"
         placeholder="Search"
-        className="pl-4 focus:border-blue-500 placeholder-gray-400"
+        className="pl-4 text-[12px] focus:border-blue-500 placeholder-gray-400"
         value={searchQuery}
         onChange={(e) => {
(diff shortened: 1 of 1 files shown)


seveibar:
@anas-sarkez screenshot?

anas-sarkez:
> @anas-sarkez screenshot?

Screenshot 2024-11-10 at 7 57 06 PM

anas-sarkez:
here if I reduced to 10px
Screenshot 2024-11-10 at 7 59 10 PM

seveibar:
Can you create a playwright test that searches for something with results.

seveibar:
I cant tell from your screenshot how results will appear because you have an empty state, a playeright test is the best way to provide visual confirmation so we should switch to those instead

anas-sarkez:
I'm not sure what are you want to reduce? the text input or search results?

Screenshot 2024-11-10 at 8 15 06 PM
Screenshot 2024-11-10 at 8 16 43 PM

seveibar:
@anas-sarkez can you create a playwright test for this? The issue isn't your code (yet), the issue is the relative size of the window, playwright tests ensure that the dimensions are known (they're better than screenshots)

It should be easy to make for this one! and will ensure future people don't break it

PR #182, opened 2024-11-11
refactor: search text

RohittCodes:
playwright tests are facing errors with the initial xs and md devices since they're hidden. will be pushing them after feedbacks (if necessary)

playwright-test for lg-devices:
before: search spec ts-Search-Input-lg

after-fix: search spec ts-Search-Input-lg-1 (2)

/claim #79 
fixes: #79


seveibar (changes requested):
i'm requiring playwright tests for this one

seveibar (approved):
looks good! but the snapshot is out of date

RohittCodes:
@seveibar updated the snapshots!

seveibar:
@RohittCodes sorry i wrote that comment super quickly and didn't explain, see how "snippets" is in the header? If you merge with main we no longer have "snippets" in the header

seveibar (approved):
going to give a pass on the snapshot issue for now because other snapshots have the same issue and it doesn't seem to impact the testing too dramatically

Issue #180, opened 2024-11-10
Click to copy doesn't work on snippets page


seveibar:
![Image](https://github.com/user-attachments/assets/706ac2b4-ece2-41b6-99b2-0046be11c53f)

ahmedhalac:
Hi @seveibar. Can I work on this?

imrishabh18:
@ahmedhalac Please go ahead and raise a PR for this

PR #184, opened 2024-11-11
fix(#180): Added copy to clipboard functionality

ahmedhalac:
This fix is resolving issue with copy to clipboard on the snippets page.

Related to #180

### DIFF
--- a/src/components/ViewSnippetSidebar.tsx
+++ b/src/components/ViewSnippetSidebar.tsx
@@ -20,30 +20,17 @@ import { cn } from "@/lib/utils"
 import { useCurrentSnippet } from "@/hooks/use-current-snippet"
 import { useToast } from "@/hooks/use-toast"
 import { useFilesDialog } from "./dialogs/files-dialog"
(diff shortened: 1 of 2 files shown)


src/components/ViewSnippetSidebar.tsx
```diff
@@ -23,10 +23,27 @@ import { useFilesDialog } from "./dialogs/files-dialog"
...
+        title: "Copied to clipboard!",
+        description: "The text has been copied successfully.",
+      });
+    } catch (error) {
+      toast({
+        title: "Failed to copy",
+        description: "There was an error copying the text.",
+      });
+    }
+  };
```
seveibar, line 46: can you put this in a separate hook called `useCopyToClipboard`

Also you'll need to run `bun format` to match the code style

seveibar (changes requested):
nice work, minor changes needed

src/components/ViewSnippetSidebar.tsx
ahmedhalac: Fixed

Issue #197, opened 2024-11-13
Add support for unstarring a snippet


seveibar:
Sometimes we want to unstar things instead of starring them

![Image](https://github.com/user-attachments/assets/90a496ba-d343-407d-9466-7f44a70d3b2b)

Note that the api is supposed to return `snippet.is_starred` to indicate that the current user starred the snippet

RohittCodes:
should i create another route for it? the initial one would be good enough to handle the add, and remove logic if it's not starred. like toggle the button to star and remove the star.

and it may not need much work on the frontend too. @seveibar

PR #202, opened 2024-11-13
feat: unstar snippets

RohittCodes:
included feature to remove the star from the snippets. this implementation maintains the initial `/add_star` route (in my opinion we could rename it to `/star`, for clear and simple implementation).

and if the implementation isn't on point, i would be willing to update it with the correct `/remove_star` if asked.

fixes: #197 
/claim #197


seveibar (changes requested):
Should be /remove _star we use “http sdk” convention, meaning routes are organized like sdk methods

bun-tests/fake-snippets-api/routes/snippets/add_star.test.ts
```diff
@@ -61,54 +61,3 @@ test("add star to non-existent snippet", async () => {
...
-        },
-      },
-    )
-    expect(true).toBe(false) // Should not reach here
-  } catch (error: any) {
-    expect(error.status).toBe(400)
-    expect(error.data.error.message).toBe(
-      "You have already starred this snippet",
-    )
-  }
```
seveibar, line 113: why remove this test?

fake-snippets-api/routes/api/snippets/add_star.ts
```diff
@@ -24,14 +24,6 @@ export default withRouteSpec({
     })
   }
 
-  // Check if already starred
-  if (ctx.db.hasStarred(ctx.auth.account_id, snippet_id)) {
-    return ctx.error(400, {
-      error_code: "already_starred",
-      message: "You have already starred this snippet",
-    })
-  }
```
seveibar, line 33: why is this removed?

fake-snippets-api/routes/api/snippets/get_star.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  // Check if snippet is already starred
+  const isStarred = ctx.db.hasStarred(ctx.auth.account_id, snippet_id)
+
+  return ctx.json({
+    ok: true,
+    is_starred: isStarred,
+  })
+})
```
seveibar, line 33: we don't need this method. `snippet.is_starred` should be added, the original issue mentioned this

fake-snippets-api/routes/api/snippets/remove_star.ts
```diff
@@ -0,0 +1,43 @@
...
+    account_id: ctx.auth.account_id,
+    snippet_id,
+    has_starred: false,
+    created_at: new Date().toISOString(),
+    updated_at: new Date().toISOString(),
+  }
+
+  return ctx.json({
+    ok: true,
+    account_snippet: accountSnippetSchema.parse(accountSnippet),
```
seveibar, line 41:
```suggestion
```

fake-snippets-api/routes/api/snippets/remove_star.ts
```diff
@@ -0,0 +1,43 @@
...
+  ctx.db.removeStar(ctx.auth.account_id, snippet_id)
+
+  // parse account snippet
+  const accountSnippet = {
+    account_id: ctx.auth.account_id,
+    snippet_id,
+    has_starred: false,
+    created_at: new Date().toISOString(),
+    updated_at: new Date().toISOString(),
+  }
```
seveibar, line 37: this isn't needed if we remove the return type

seveibar (changes requested):
see comments

fake-snippets-api/routes/api/snippets/add_star.ts
RohittCodes: ig there was no need to include an already starred condition as we were handling it in the `/remove_star` route. will revert these changes.

fake-snippets-api/routes/api/snippets/remove_star.ts
RohittCodes: should i just return `snippet.is_starred` with an ok response here?

src/components/ViewSnippetHeader.tsx
```diff
@@ -62,6 +62,8 @@ export default function ViewSnippetHeader() {
     },
   })
 
+  const [isStarred, setIsStarred] = useState(false)
```
seveibar, line 65: shouldn't you use `snippet.is_starred` here?

seveibar (commented):
looking good, one comment

fake-snippets-api/routes/api/snippets/remove_star.ts
seveibar: The snippet from `/snippets/get` should have `snippet.is_starred` based on the user

seveibar: when you star it, `qc.invalidQueries` is used to make `/snippets/get` repull the latest version of the snippet including the star status

src/components/ViewSnippetHeader.tsx
RohittCodes: yeah, but it'll trigger the unstar method, and will eventually unstar the snippet even before the user wants it to.

ig, I have to remove this hook for now.

seveibar: `/snippets/get` returns `snippet.is_starred`. Not sure what you mean by it'll trigger the unstar method

seveibar: If `/snippets/get` doesn't return `snippet.is_starred` with the specific property set for the user, then that has to be implemented as well

RohittCodes: ohh, okay. so it means i have to `/snippets/get` too. and use it inside the isStarred hook.

i did it earlier, but u asked to remove that method

seveibar: idk what the isStarred hook is. You don't need to make any new hooks afaik

when you run `qc.invalidateQueries(["snippets", snippet!.snippet_id])` it automatically causes the snippet to be reloaded

RohittCodes: yeah i've removed it now. it was from one of my previous commits, and i forgot to do it in the past commit.

RohittCodes: @seveibar can u look into this comment and just update me

seveibar: yes you don't need `useState` or `isStarred`, in fact it is bad to have it because it confuses the source of truth of whether or not the snippet is starred

seveibar (approved):
i'd recommend adding a playwright test to make sure nobody breaks this functionality in the future but this is acceptable, nice work

RohittCodes:
@seveibar done with playwright tests, i don;t know what's the issue with `bun test` check, it passes all the 34 and skips 1 locally. and i haven't changed any api routes this time.

seveibar (changes requested):
this is close but if the snippet is starred, you should say "Starred" and have a filled-in star 
image

RohittCodes:
okay, so can i create a method to get the is_starred from `snippets/get_star` and store it inside a variable using `useEffect`, and `useState` hooks (they'll be required now).

fake-snippets-api/routes/api/snippets/get_star.ts
```diff
@@ -0,0 +1,39 @@
...
+      ok: true,
+      is_starred: true,
+    })
+  }
+
+  return ctx.json({
+    ok: true,
+    is_starred: false,
+  })
+})
```
seveibar, line 39: remove this route

src/components/ViewSnippetHeader.tsx
```diff
@@ -55,6 +55,23 @@ export default function ViewSnippetHeader() {
...
+        const { data } = await axios.post("/snippets/get_star", {
+          snippet_id: snippet.snippet_id,
+        })
+        setIsStarred(data.is_starred)
+      } catch (error: any) {
+        console.error("Error fetching star status:", error)
+      }
+    }
+    fetchStarStatus()
+  }, [snippet])
```
seveibar, line 73: You need to add support for `snippet.is_starred`, the production api already has this so no other implementation is acceptable

seveibar (changes requested):

seveibar:
@RohittCodes `snippets/get` returns `snippet.is_starred` in the production api. The fake api should update to reflect this. `/snippets/get_star` should is not acceptable for the implementation

RohittCodes:
i've tried it, but there is no way to get star status if someone has, actually starred it, as according to the current setup, `auth: "none",` in `snippets/get`

if i change it to something like this: `auth: "session",`
then users who are not logged in will face issues with it.

can u just share me how is the logic setup in the prod env.

seveibar:
@RohittCodes you have to introduce a new auth called "optional_session" that adds ctx.auth only if there's a token

seveibar:
image

Something like this

RohittCodes:
Okhh got it.

seveibar (approved):
Nice work

RohittCodes:
do we need any rework on this? @seveibar

PR #208, opened 2024-11-14
feat: Download the dsn file

imrishabh18:
Screenshot 2024-11-14 at 17 41 52

Currently, the dsn free routing and the circuit-to-svg are in correct orientation. While the pcb-viewer is vertically flipped.


seveibar (approved):
To verify which is flipped, we should set pcbY to 2 or some positive number and see which goes to the right position

Issue #171, opened 2024-11-06
Allow iframe embedding of snippets


seveibar:
Will require serving PreviewContent from a URL

the route could be something like `/preview?snippet_id=...`

No other page context except PreviewContent, automatically scaled to iframe size

RohittCodes:
@seveibar if I understand correctly, this should be a separate page for previews only. like there is a `copy url` button for sharing editor, which routes to `/editor?snippet_id=...`

![Image](https://github.com/user-attachments/assets/8bcb1bb0-e2de-4181-83c9-e69b6eb87918)

If so, can I provide a dropdown here, instead of copy url, with two separate buttons for share preview and share code/editor?

seveibar:
@RohittCodes from the "view snippets" page, users can get the iframe url. I'd prefer not to change that button on the editor page for now, I have some other plans for it

PR #236, opened 2024-11-16
Update link for iframe embedding of snippets

mrudulpatil18:
continuation of #171

updated embed link with /preview

image

cc: @seveibar


seveibar (approved):
nice work!

PR #246, opened 2024-11-17
feature: insert chip

### LINKED ISSUE #172: Dialog for inserting a footprinter string
Allows selecting footprint etc. seeing all available footprints, customize parameters

Insert dropdown allows selecting "Footprinter String" (has no other options initially)

Image

Image

ShiboSoftwareDev:
There is still work that needs to be done in the footprinter repo, but the feature is mostly working.

/claim #172 
/closes #172 
image


src/components/FootprintDialog.tsx
```diff
@@ -0,0 +1,361 @@
...
+                onOpenChange(false)
+              }}
+              disabled={!footprinterString || !chipName}
+              className="w-full"
+            >
+              Insert Footprint
+            </Button>
+          </div>
+          <div className="rounded-xl overflow-hidden w-[800px] h-[600px]">
+            {previewSvg && <div>{parse(previewSvg)}</div>}
```
seveibar, line 355: You can just use `_dangerouslySetInnterHtml`- this is probably more efficient than adding an expensive dependency

seveibar (approved):
@ShiboSoftwareDev looks good, I would recommend adding a snapshot test that does something basic because otherwise people will probably break this functionality.

The dependency you added shouldn't be necessary, we already have a really big bundle so I recommend removing it

Great work!!!!

ShiboSoftwareDev:
@seveibar can you review this? I did some playwright tests. in under md the dialog component needs to be scrolled inside but the website as a whole does too. Do I have to try to make it responsive? It would be difficult with the svg generated having fixed dimensions.

seveibar (approved):
fantastic work @ShiboSoftwareDev , we will make responsiveness a future issue.

In general, we want everything to be responsive but we can do it after merging desktop versions to keep PRs from getting too big. The main reason is i frequently use my phone and importantly demo to investors using my phone!

PR #252, opened 2024-11-20
Revert: cursor pos

seveibar:
Reverts tscircuit/snippets#250

@RohittCodes you caused a regression where every snippet has red underlines until you modify it

image


RohittCodes:
Okay, will go through the issue again. can u open it, @seveibar or should I raise a pr directly??

Issue #265, opened 2024-11-20
"Unsaved Changes" is always showing, even after saving


seveibar:
I think @RohittCodes may have broken this, @RohittCodes can you fix?

![Image](https://github.com/user-attachments/assets/8435ba25-b1a0-485c-b68f-a66fca5f61a5)

PR #268, opened 2024-11-21
fix: load manual_json

RohittCodes:
`manualEditsFileContent` wasn;t being initialized with the manual_edit_json when the page loaded, so the issue was when we tried to look for `hasUnsavedChanges` it always resulted in true, and that's the reason it displayed "Unsaved Changes"

fixes: #265 
sorry for this bug too. cc: @seveibar

### DIFF
--- a/src/components/CodeAndPreview.tsx
+++ b/src/components/CodeAndPreview.tsx
@@ -41,12 +41,12 @@ export function CodeAndPreview({ snippet }: Props) {
     )
   }, [])
 
-  // Initialize with template or snippet's manual edits if available
-  const [manualEditsFileContent, setManualEditsFileContent] = useState(
-    snippet?.manual_edits_json ??
-      JSON.stringify(manualEditsTemplate, null, 2) ??
-      "",
-  )
+    // Initialize with template or snippet's manual edits if available
+    const [manualEditsFileContent, setManualEditsFileContent] = useState(
+      snippet?.manual_edits_json ??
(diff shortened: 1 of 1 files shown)


src/components/CodeAndPreview.tsx
```diff
@@ -64,6 +64,13 @@ export function CodeAndPreview({ snippet }: Props) {
 
   const { toast } = useToast()
 
+  useEffect(() => {
+    if (snippet?.manual_edits_json) {
+      console.log('Setting manual edits from snippet:', snippet.manual_edits_json);
+      setManualEditsFileContent(snippet.manual_edits_json);
+    }
+  }, [snippet?.manual_edits_json])
```
seveibar, line 72: you only want to do this the first time it loads

```suggestion
  }, [Boolean(snippet?.manual_edits_json)])
```

src/components/CodeAndPreview.tsx
```diff
@@ -64,6 +64,13 @@ export function CodeAndPreview({ snippet }: Props) {
 
   const { toast } = useToast()
 
+  useEffect(() => {
+    if (snippet?.manual_edits_json) {
+      console.log('Setting manual edits from snippet:', snippet.manual_edits_json);
```
seveibar:
```suggestion
```

seveibar (changes requested):
see comments

seveibar (approved):

Issue #261, opened 2024-11-20
Create a Playwright Test that tests saving manual edits and viewing the changes inside the "View Files" menu


seveibar:
/bounty $10

CC @RohittCodes did some related work recently

PR #269, opened 2024-11-21
fix: add manual_json validation and playwright-test

RohittCodes:
refactored: manual_edit_json template ingestion into db.
included playwright tests

fixes: #261 
/claim #261

cc: @seveibar


RohittCodes:
forgot to claim this earlier
/claim #265

RohittCodes:
i forgot to claim the bounty for #265. and the fix for it temporarily solved the issue. i.e., whenever a new snippet was created it still showed the "Unsaved changes" text. This PR solves the issues with `manual_edit_json`.

fake-snippets-api/lib/db/db-client.ts
```diff
@@ -88,8 +88,15 @@ const initializer = combine(databaseSchema.parse({}), (set, get) => ({
...
     const newSnippetId = `snippet_${get().idCounter + 1}`
+    const manualEditsJsonTemplate = {
+      pcb_placements: [],
+      edit_events: [],
+      manual_trace_hints: [],
+    }
+
     const newSnippet = snippetSchema.parse({
       ...snippet,
+      manual_edits_json: JSON.stringify(manualEditsJsonTemplate, null, 2),
```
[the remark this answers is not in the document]
seveibar: if there is an issue with the default value, that should be handled on the frontend the first time it's selected

seveibar (changes requested):
you didn't test saving the manual edits file, you just set a default value on the server and showed it loads in both views.

you must hit run/save and ensure that edits to it are saved

fake-snippets-api/lib/db/db-client.ts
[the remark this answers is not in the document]
RohittCodes: Uhh, sure..

RohittCodes:
@seveibar all the failing tests are passing in my local setup again. Not sure what's causing the checks fail here. 🤷‍♂️

seveibar:
@RohittCodes can you make sure youve got the latest deps by running "bun i" locally then rerunning tests. Is it a timeout issue on ci?

RohittCodes:
yeah all the deps are up to date, and i don't think there is any timeout issue for `bun tests`, but the playwright-test has timeout issues on `preview-page`. @seveibar

seveibar:
Hmm maybe configure for longer timeout in the config/workflow?

RohittCodes:
@seveibar nah this too doesn't work.

src/components/CodeAndPreview.tsx
```diff
@@ -41,12 +42,22 @@ export function CodeAndPreview({ snippet }: Props) {
...
+        JSON.stringify(manualEditsTemplate, null, 2)
+      // Validate that it's parseable JSON
+      JSON.parse(initialContent)
+      return initialContent
+    } catch (e) {
+      console.warn(
+        "Invalid initial manual edits content, using default template",
+      )
+      return JSON.stringify(manualEditsTemplate, null, 2)
+    }
```
seveibar, line 59: don't have redundant logic for the default manual edits json, the `useEffect` already handles it

src/components/CodeAndPreview.tsx
```diff
@@ -62,20 +73,38 @@ export function CodeAndPreview({ snippet }: Props) {
...
 
-  const { toast } = useToast()
-
+  // Update manual edits when snippet changes, with validation
   useEffect(() => {
     if (snippet?.manual_edits_json) {
-      setManualEditsFileContent(snippet.manual_edits_json)
+      try {
+        JSON.parse(snippet.manual_edits_json)
+        setManualEditsFileContent(snippet.manual_edits_json)
```
seveibar, line 81: this should be returned as JSON from the api, if it's not we should make it so it does

src/components/CodeAndPreview.tsx
```diff
@@ -62,20 +73,38 @@ export function CodeAndPreview({ snippet }: Props) {
...
+      return {
+        "./manual-edits.json": JSON.parse(manualEditsFileContent),
+      }
+    } catch (e) {
+      console.warn("Error parsing manual edits for imports, using empty object")
+      return {
+        "./manual-edits.json": {},
+      }
+    }
+  }, [manualEditsFileContent])
```
seveibar, line 107: this try catch pattern is not good, let's revert that

src/components/CodeAndPreview.tsx
seveibar: if it's returned as a string, the name should be `manual_edits_json_content`

src/components/CodeAndPreview.tsx
```diff
@@ -101,6 +130,14 @@ export function CodeAndPreview({ snippet }: Props) {
   const updateSnippetMutation = useMutation({
     mutationFn: async () => {
       if (!snippet) throw new Error("No snippet to update")
+
+      // Validate manual edits before sending
+      try {
+        JSON.parse(manualEditsFileContent)
+      } catch (e) {
+        throw new Error("Invalid manual edits JSON")
+      }
```
seveibar, line 139: this is good

src/components/CodeAndPreview.tsx
seveibar: actually this is ok, but this file is getting too big so create a utility function in the utils directory `parseJsonOrDefault(manualEditsFileContent, {})`

seveibar (changes requested):
this is a pretty good implementation, but i'm very critical of this file getting too big and difficult to understand so I'm going to insist on some small details, mostly...

- Only handle setting manual edits json in a single point
- We can remove any code that uses the template, we don't need the template especially if we set it to the empty project when it fails to parse
- The failure to parse really shouldn't go to an empty object- that basically means that we're overriding the user data when we fail to parse and erasing the previous content- this is very bad and buggy behavior. We should instead set the manualEditsFileContent to the string, then allow the user to edit it to correct it, not erase it

playwright-tests/manual-edits.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  await filesLink.waitFor({ state: "visible" })
+  await filesLink.click()
+
+  const fileLink = page.getByText("manual-edits.json")
+  await fileLink.waitFor({ state: "visible" })
+  await fileLink.click()
+
+  await page.waitForLoadState("networkidle")
+  await expect(page).toHaveScreenshot("manual-edits-view.png")
+})
```
seveibar, line 80: this is great except i'm not sure why the screenshot is incorrect, the component hasn't moved

src/lib/utils/parseJson.ts
```diff
@@ -0,0 +1,8 @@
+export const parseJsonOrString = (jsonString: string): Record<string, any> => {
+  try {
+    return JSON.parse(jsonString)
+  } catch (e) {
+    // Convert string to an object with a value property if parsing fails
+    return { value: jsonString }
```
seveibar, line 6: this is a bad pattern,

1) we never use the meaningless variable name `value` unless there is some industry-specific reason we have to

2) the function name makes what this does unclear

3) if something fails to parse, there are only two appropriate conventions: return null OR return a "go-style" error e.g. `return { error } | { error: null, json }`

4) the filename doesn't match the function name

src/components/CodeAndPreview.tsx
```diff
@@ -41,12 +41,7 @@ export function CodeAndPreview({ snippet }: Props) {
     )
   }, [])
 
-  // Initialize with template or snippet's manual edits if available
-  const [manualEditsFileContent, setManualEditsFileContent] = useState(
-    snippet?.manual_edits_json ??
-      JSON.stringify(manualEditsTemplate, null, 2) ??
-      "",
-  )
+  const [manualEditsFileContent, setManualEditsFileContent] = useState("{}")
```
seveibar, line 44: you're not consistent with the default value- make it the empty string as it is in other places

src/components/CodeAndPreview.tsx
```diff
@@ -65,14 +60,14 @@ export function CodeAndPreview({ snippet }: Props) {
...
+    if (snippet?.manual_edits_json_content) {
+      setManualEditsFileContent(snippet.manual_edits_json_content)
     }
-  }, [Boolean(snippet?.manual_edits_json)])
+  }, [Boolean(snippet?.manual_edits_json_content)])
 
   const userImports = useMemo(
     () => ({
-      "./manual-edits.json": JSON.parse(manualEditsFileContent),
+      "./manual-edits.json": parseJsonOrString(manualEditsFileContent),
```
seveibar, line 70:
```suggestion
      "./manual-edits.json": parseJsonOrNull(manualEditsFileContent) ?? "",
```

seveibar (changes requested):
really really close, most of the issue is with the unconventional function. After that, we just need to consider why the screenshot is wrong- is the functionality broken? If not, why is the screenshot incorrect?

src/components/CodeAndPreview.tsx
```diff
@@ -43,9 +43,7 @@ export function CodeAndPreview({ snippet }: Props) {
 
   // Initialize with template or snippet's manual edits if available
   const [manualEditsFileContent, setManualEditsFileContent] = useState(
-    snippet?.manual_edits_json ??
-      JSON.stringify(manualEditsTemplate, null, 2) ??
-      "",
+    parseJsonOrNull(""),
```
seveibar, line 46: Just say null

src/components/CodeAndPreview.tsx
```diff
@@ -101,13 +99,21 @@ export function CodeAndPreview({ snippet }: Props) {
   const updateSnippetMutation = useMutation({
     mutationFn: async () => {
       if (!snippet) throw new Error("No snippet to update")
+
+      // Validate manual edits before sending
+      try {
+        parseJsonOrNull(manualEditsFileContent)
```
seveibar, line 105: praseJsonOrNull cant throw, remove this try catch

seveibar (changes requested):
We’re very close now

seveibar (approved):
Cc @imrishabh18 we need to update the prod api to support the column rename

seveibar:
@RohittCodes I looked at the failing snapshots and they look like something is legit broken:

image

RohittCodes:
yeah will look into this 👍

RohittCodes:
the issue was with the imports in view-snippet. fixed it! and it seems like there were changes on the editor page, will update the playwright tests for that.
Screenshot 2024-12-09 135957

https://github.com/user-attachments/assets/b0000806-214e-41d1-83c7-3bc78c1656e8

seveibar:
@RohittCodes nice work

Issue #253, opened 2024-11-20
Trigger the editor to reevaluate types after files load without changing the cursor position


seveibar:
Related to: https://github.com/tscircuit/snippets/pull/252

/bounty $7

PR #270, opened 2024-11-21
fix: cursor-pos

RohittCodes:
this fixes the exact cursor position, without reflecting the code, and its types.

/claim #253
fixes: #253


seveibar (approved):
Good fix

PR #271, opened 2024-11-21
added size analysis workflow

ShiboSoftwareDev:


seveibar:
@ShiboSoftwareDev waiting for the comment still but i think you want to use a vite specific analysis, i dont think the chunk sizes are meaningful relative to a vite/rollup bundle analysis with size attribution for imports etc

ShiboSoftwareDev:
> but i think you want to use a vite specific analysis, i dont think the chunk sizes are meaningful relative to a vite/rollup bundle analysis with size attribution for imports etc

I think I get it

ShiboSoftwareDev:
@seveibar the workflow is not allowed to comment

seveibar:
> @seveibar the workflow is not allowed to comment

Hmm i wonder how we can fix this for forks… we have this issue on core too

ShiboSoftwareDev:
@seveibar resource not accessible by integration. maybe there is something you need to enable in settings

PR #286, opened 2024-11-25
size analysis workflow

### LINKED ISSUE #259: Add bundle size analysis comment as a comment on each pull request
I want to start tracking our bundle size. Create a github workflow that comments on each PR with the bundle size and gives a breakdown of the size. We have one in [tscircuit/core](https://github.com/tscircuit/core) but it's not purpose-built for vite applications.

/bounty $40

ShiboSoftwareDev:
/claim #259 
/closes #259


ShiboSoftwareDev:
finally got it to show, now need to see why the analysis is not showing anything

PR #291, opened 2024-11-26
bundle-analysis

ShiboSoftwareDev:
/claim #259


seveibar:
For some reason the top dependencies dont display for me? Is it still WIP?

Fwiw generally if a bash script gets long or has custom parsing, its much better to create a scripts file that runs a parser

seveibar (approved):
nice work

Issue #295, opened 2024-11-27
when trying to rename a snippet and fail the dialog screen freez until you remove it , it should allow you to try another name


Abse2001:
https://github.com/user-attachments/assets/bc6a89ce-1cab-4c0b-a147-b599136329b5

PR #297, opened 2024-11-28
fix: rename timeout

RohittCodes:
added a timeout and introduced a finally block

fixes: #295
/claim #295

### DIFF
--- a/src/components/dialogs/rename-snippet-dialog.tsx
+++ b/src/components/dialogs/rename-snippet-dialog.tsx
@@ -27,15 +27,18 @@ export const RenameSnippetDialog = ({
   const [pending, setPending] = useState(false)
 
   const handleRename = async () => {
+    // Prevent multiple submissions while pending
+    if (pending) return
     try {
       setPending(true)
+      // timeout to prevent rapid calls
+      await new Promise((resolve) => setTimeout(resolve, 500))
       await axios.post("/snippets/update", {
         snippet_id: snippetId,
         unscoped_name: newName,
       })
       onRename?.(newName)
       onOpenChange(false)
-      setPending(false)
       toast({
         title: "Snippet renamed",
         description: `Successfully renamed to "${newName}"`,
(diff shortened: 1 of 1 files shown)


seveibar (changes requested):
this is fine, but can you switch this to use a react-query mutation instead? This type of logic should always use react-query

RohittCodes:
this an api endpoint that failed. no changes made from my end.

seveibar:
@RohittCodes use this

image


@Abse2001 can you confirm this fixes the underlying issue? I'm not super familiar w/ the problem

@RohittCodes nice refactor

Abse2001:
@seveibar I will need few hours before I can test it Im going to collage.

Abse2001:
It happens when you try to use a name that is not acceptable, eg (using spaces), after the fail message the dialog screen just stuck on renaming state, you can't change the name until you close it and open a new dialog.

Abse2001:
@RohittCodes sorry for the delay I tested it and it looks fixed.
just fix the bun test and you are all good.

RohittCodes:
@Abse2001 I actually don't know why the bun tests failed, I've not even tried changing anything in the api.

It's completely a frontend change.

seveibar:
I think the bun tests are kind of flaky, just gotta rerun them (if that is the cause)

src/components/dialogs/rename-snippet-dialog.tsx
```diff
@@ -24,31 +24,39 @@ export const RenameSnippetDialog = ({
...
         title: "Error",
         description: "Failed to rename the snippet. Please try again.",
         variant: "destructive",
       })
-    }
+    },
+  })
+
+  const handleRename = () => {
+    renameSnippetMutation.mutate()
```
seveibar, line 59: Just use an inline function here, these “handle” functions rarely increase readability especially when they are just proxying a mutation

seveibar (approved):

PR #302, opened 2024-11-29
fix:regex to import names with dash properly

### LINKED ISSUE #301: Import names with dash dont properly import
image

Anshgrover23:
/claim #301
Fixes #301 

Screenshots:
image

### DIFF
--- a/src/lib/__tests__/constants.test.ts
+++ b/src/lib/__tests__/constants.test.ts
@@ -30,7 +30,7 @@ describe("TSCI_PACKAGE_PATTERN", () => {
     "@tsci/123start-with-number", // Starts with number
     "@tsci/-start-with-dash", // Starts with dash
     "@tsci/.start-with-dot", // Starts with dot
-    
+
     "@tsci/double..dot", // Double dot
     "@tsci/end-with-dash-", // Ends with dash
     "@tsci/end-with-dot.", // Ends with dot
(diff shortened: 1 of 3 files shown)


seveibar (changes requested):
Centralize regex (define exactly once) and prove it works inside bun unit tests. Test with packages with multiple dashes

src/lib/__tests__/constants.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    })
+  })
+
+  // Test invalid patterns
+  const invalidPackages = [
+    "@tsci/", // Empty
+    "@tsci/123start-with-number", // Starts with number
+    "@tsci/-start-with-dash", // Starts with dash
+    "@tsci/.start-with-dot", // Starts with dot
+    "@tsci/double--dash", // Double dash
```
seveibar, line 32: Double dash should be acceptable

src/lib/__tests__/constants.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    "@tsci/double..dot", // Double dot
+    "@tsci/end-with-dash-", // Ends with dash
+    "@tsci/end-with-dot.", // Ends with dot
+    "@tsci/invalid@chars", // Invalid characters
+    "@tsci/dot.dash.-mixed", // Invalid dot-dash combination
+    "@tsci/dash-.dot-mixed", // Invalid dash-dot combination
```
seveibar, line 38: This is also acceptable although you dont need to fix it in this PR because its rare

seveibar (approved):
Good work, please address comment, but the work is well done so acceptable as an improvement as is so approving

excellent job

seveibar (approved):

PR #307, opened 2024-11-30
refactor: search-links

### LINKED ISSUE #104: When doing a search for a component, only open in a new tab if you're on the `/editor` or `/ai` page (where you might have changes you want to save)

RohittCodes:
fixes #104


src/components/SearchComponent.tsx
```diff
@@ -52,6 +54,10 @@ const SearchComponent: React.FC<SearchComponentProps> = ({
     }
   }, [])
 
+  const shouldOpenInNewTab = () => {
```
seveibar, line 57: This should be a variable not a function

seveibar (changes requested):
Is there a way to do this that uses the Link element

seveibar:
A lot of the code is repeated in the ternary, why not just have the prop to Link change? Does it not work?

RohittCodes:
Link component from the wouter library doesn't handle `targets` natively, it loads the content in the same tab as part of SPA navigation. I’ve tried using anchor tags too, but I noticed that they trigger a full-page reload unless handled by the router so came up with this @seveibar.

seveibar:
@RohittCodes ok, if you fix the var/func issue then ill approve

src/components/SearchComponent.tsx
```diff
@@ -52,6 +54,8 @@ const SearchComponent: React.FC<SearchComponentProps> = ({
     }
   }, [])
 
+  const openInNewTab = location === "/editor" || location === "/ai"
```
seveibar, line 57: don't use verbs for variable names, verbs are for function names. `shouldOpenInNewTab` is a var name

```suggestion
  const shouldOpenInNewTab = location === "/editor" || location === "/ai"
```

seveibar (changes requested):

seveibar (changes requested):
can you add a test for this, sorry it's taking so long but i'm skeptical of the location only having the classname and not query parameters etc.

RohittCodes:
@seveibar should i work on the existing search specs for the tests or should i create a new file?

seveibar:
I generally prefer new files!

playwright-tests/search-links.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const popupPromise = page.waitForEvent("popup")
+  await page.getByRole("link", { name: "testuser/my-test-board A" }).click()
+
+  const popup = await popupPromise
+
+  await popup.waitForLoadState("networkidle")
+
+  const popupUrl = popup.url()
+  console.log(popupUrl)
```
seveibar, line 26:
```suggestion
```

seveibar (approved):

Issue #315, opened 2024-12-01
Playwright Test for "Cmd Click to go to snippet"


seveibar:
This functionality keeps breaking and it's driving me nuts- it seriously impacts DX. Let's make a test for it

/bounty $20

Dependent on it actually being fixed #313

seveibar:
doesn't work on this snippet: https://tscircuit.com/editor?snippet_id=48b2b687-2e72-4d2e-8a1e-5afafb0b11ba

seveibar:
woops wrong issue

PR #318, opened 2024-12-01
test for underlined and clickable imports

mrudulpatil18:
Issue: #315 

Added a playwright test for:
- checking underlined imports
- ctrl/cmd clickable imports
- snapshot test

/claim #315 

cc: @seveibar


seveibar (approved):
nicely done

Issue #324, opened 2024-12-01
Schematic and PCB View should expand to full size when view is expanded


seveibar:
![Image](https://github.com/user-attachments/assets/8ea74ce0-bb19-42d9-9b4b-df13fed473ca)

Bonus points/tip if you can also fix the schematic view never expanding fully even when not expanded

![Image](https://github.com/user-attachments/assets/c821bee1-8300-44bb-95c3-40888f20fa32)

/bounty $15

PR #327, opened 2024-12-02
feat: expand-to-full

RohittCodes:
fixes: #324 
/claim #324 
Image
image

cc: @seveibar


RohittCodes:
@seveibar i guess for the pcb height, it needs some changes in the component itself, it handles the height in its own way unlike others

RohittCodes:
@seveibar can you re-run the bun tests? i don't see any options on my side. and should i add a function or variable for the styles it's increasing the file size and is being repeated.

RohittCodes:
yeah, it did!! shibo did some fix in the features for the `bun-tests`

seveibar (approved):

PR #330, opened 2024-12-02
added static skeleton page with fixed data

### LINKED ISSUE #323: Show "skeleton page" when waiting for view snippet page
Currently we show "Loading" when waiting for a page to load

A skeleton page is a version of a page where it is structurally laid out correctly while details wait to load.

The snippet name should be shown (it is accessible in a variety of ways, including the react query cache and url path name)

Please keep the solution clean, read over your solution and make sure you're not missing simpler solutions.

https://github.com/user-attachments/assets/35d0c356-a1d5-4f1e-b0b0-0b7d9a807bb1

/bounty $30

mrudulpatil18:
fixes #323 

To keep things simple, created new static components for loading view.
these components mimic the normal ones and look the same.

This is how it looks in loading mode

https://github.com/user-attachments/assets/9aa6c50c-766b-415f-a309-bbdfc1291593

/claim #323 

cc: @seveibar

### DIFF
--- a/src/components/StaticPreviewContent.tsx
+++ b/src/components/StaticPreviewContent.tsx
@@ -27,19 +27,19 @@ export default function StaticPreviewLoading() {
       <div className="md:sticky md:top-2">
         <Tabs defaultValue="code" className="flex-grow flex flex-col">
           <div className="p-4 border-b border-gray-200 flex items-center gap-2">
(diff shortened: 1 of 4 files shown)


seveibar:
this is pretty impressive. it looks great!

PR #339, opened 2024-12-03
auto-run code

### LINKED ISSUE #334: When editor page first loads, use the `circuit_json` from the snippet so they user doesn't need to click `Run`
For both editor and view snippets page

/bounty $20

ShiboSoftwareDev:
is this specific to entering a snippet? Because in this implementation whenever you enter any editor it automatically runs the code
/claim #334
/closes #334


seveibar (changes requested):
We dont want to autorun code because 1) it freezes browsers atm on some snippets 2) insecure code loading and 3) the precompiled circuit json is already available and saved on the snippet, ready for display

ShiboSoftwareDev:
I moved the logic to view-snippet now it only runs for snippets

seveibar (changes requested):
we don't want to run snippets on page load, the circuit json is already compiled.

Think of it this way- it takes 10 seconds to autoroute, why bother autorouting circuit json when there's already circuit json available on `snippet.circuit_json`?

ShiboSoftwareDev:
Oooh, got it

ShiboSoftwareDev:
https://github.com/user-attachments/assets/d879353d-fa80-4f2d-8f4b-65e6280901a5

seveibar (changes requested):
i don't mind this solution but it violates one-way-data-flow- there's a `circuitJson` prop passed to `<PreviewContent />`, so you should just have some logic to set that

having `useEffect` all over the place eventually makes a codebase unusable, so you want it's usage to have pretty local effects (only interact with the immediately local or session state machines unless absolutely necessary)

```tsx
const circuitJsonForPreview = tsxResult?.circuitJson ?? snippet?.circuit_json
const circuitJsonKeyForPreview = tsxResult?.circuitJsonKey ?? (snippet?.circuit_json ? "snippet" : "")
```

ShiboSoftwareDev:
thank you for the helpful reviews, clean code is an essential skill but I guess you can only learn to write clean code by writing bad code first

seveibar (approved):

PR #340, opened 2024-12-03
Configure TypeScript and Babel standalone to load separately from the bundle (via <script async>)

### LINKED ISSUE #335: Configure typescript and babel standalone to load separate from bundle (via `<script async`)
This will allow the snippets page to load much faster (will reduce bundle size)

Whenever typescript or babel standalone is used, you should pull it from the global `window` object (or wherever the script globally loads to)

In places where we need typescript/babel standalone available, throw an error or wait until it's loaded

The bundle size should be smaller on the PR

There may be some dependencies that need to be adjusted.

Note: This work REQUIRES #288 

/bounty $30

techmannih:
close #335 
/claim #335


seveibar:
this doesn't fix the bundle size

PR #341, opened 2024-12-03
increased playwright timeout

RohittCodes:
cc: @imrishabh18

### DIFF
--- a/.github/workflows/bun-test.yml
+++ b/.github/workflows/bun-test.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/bun-test.yml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 5
```
imrishabh18, line 13: The timeout issue is for just playwright, right? Do we need to increase it for bun tests?

RohittCodes: umm yeah, refactored it!

PR #349, opened 2024-12-05
fix: flaky tests

### LINKED ISSUE #346: Make playwright tests not flaky
I keep seeing broken playwright tests in PRs

I don't know why it's flaky, but we shouldn't be seeing that so often in PRs

/bounty $5

Anshgrover23:
Fixes #346 
/claim #346

### Description:
- I have Ran tests more than 5 times after setting this global timeout of 60 seconds.
image

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,6 +1,7 @@
 import { defineConfig } from "@playwright/test"
 
 export default defineConfig({
+  timeout: 60000,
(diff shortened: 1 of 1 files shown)


Anshgrover23:
@seveibar Done, but i think the permanent solution is we have to split the long running tests into different files, as you can see here:
image
### Yours opinion?

PR #350, opened 2024-12-05
add retries for flaky test

### LINKED ISSUE #346: Make playwright tests not flaky

techmannih:
close #346

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -2,7 +2,7 @@ import { defineConfig } from "@playwright/test"
 
 export default defineConfig({
   // Run your local dev server before starting the tests
-  retries: 5,
+  retries: process.env.CI ? 5 : 1,
   webServer: {
     command: "npm run start:playwright-server",
     url: "http://127.0.0.1:5177",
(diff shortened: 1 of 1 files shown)


playwright.config.ts
```diff
@@ -2,6 +2,7 @@ import { defineConfig } from "@playwright/test"
 
 export default defineConfig({
   // Run your local dev server before starting the tests
+  retries: 5,
```
seveibar, line 5: Can you only configure retries on ci

seveibar: Use process.env.CI ? 5 : 1

Anshgrover23:
@seveibar i think using retries can increased overall test suite time: If tests are retried frequently.

Anshgrover23:
image

## Also You can see it is retrying one more time even after tests are pass.

seveibar (approved):
@Anshgrover23 nice investigation but 2.2m vs 2.1m seems like noise not a notable difference, i could be wrong but increasing timeout rarely improves test times

Issue #342, opened 2024-12-03
Bad error message on using the platedhole footprint with shape `oval`


imrishabh18:
Bad error message, don't know what does this `x` mean here.

https://snippets.tscircuit.com/editor?&snippet_type=board#data:application/gzip;base64,H4sIAPVfT2cAA21Qvc7CMAzc+xRWJli+kk8MDE1nZhYGxBCoIZHaJErNj1Tx7oQk5UfgzXfn053x6qwnaPAgTy3BZAqihkkBUO2s9A1cdENKMD7rOgYK9VFR3uogCjKPve7J+rgBGNmhYCvO8p5oafYoBj67ZfRgLTmvDYkhI8HpCSbnjLpWEjbKtviGAjxSL4O2F8OGOW04247maXolXUhiz7Ids6TRxqBfx1oD/7yxJ3pRf/NPMt4t0wcCu/hxOrL/71z5qlOVXx2TMmqqMr68LqbFHeIZutqWAQAA

![Image](https://github.com/user-attachments/assets/8df698c8-bb3f-4799-b21e-9b0de8709360)

This error comes when you add `portHints` to the `platedhole`

PR #362, opened 2024-12-08
refactor: error-msg

RohittCodes:
Screenshot 2024-12-08 115921

the issue is most likely to be in the core package. where there is a type mismatch or some unexpected property access for the dimension x in the `_getPcbCircuitJsonBounds()`, `doInitialPcbPrimitiveRender()` methods. cc: @imrishabh18 

/claim #342 
fixes: #342

### DIFF
--- a/src/hooks/use-run-tsx/index.tsx
+++ b/src/hooks/use-run-tsx/index.tsx
@@ -205,7 +205,7 @@ export const useRunTsx = ({
           console.error("Evaluation error:", error)
           setTsxResult({
             compiledModule: module,
-            message: `Render Error: ${error.message}`,
+            message: `Render Error: ${error.message}\n\n${error.stack}`,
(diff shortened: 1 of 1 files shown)


imrishabh18 (approved):

seveibar (changes requested):
This error isnt more helpful, give a readable string that tells the user how to fix the bug (something about the shape or something)

reproduce rishabhs example in a test and test for the new error message

RohittCodes:
yeah but the error seems to be in the package but not the code. and it's fixed now. @seveibar

seveibar:
We can transfer the issue or when you update the core dep you can claim on that issue

imrishabh18:
Yeah, I have fixed this in core. After that it didn't made sense to change the message here. Closing the issue

Issue #370, opened 2024-12-10
If `manual-edits.json` has any content, show an error in the code editor if the code does not contain `"manualEdits={"`


seveibar:
- Manual edits are made by moving components on the PCB Viewer
- A user must import their manual edits file and set it as a prop in order for the manual edits to be applied
- If the user has not set the prop, but has made manual edits (the json file is not empty) then we should display an error dropdown
- We may have more errors in the future, so this is a dropdown. Hide the dropdown when there are no errors
- The dropdown should contain a helpful error message explaining what to do. If the user presses the dropdown error, we attempt to automatically fix the issue using a regular expression (searching for `<board` or `<group subcircuit` and adding the prop. If we're unable to fix it automatically, we throw a toast error

See an example of manual edits in use here: https://tscircuit.com/seveibar/rp2040-module

![Image](https://github.com/user-attachments/assets/9dc6ab3d-e6b5-4b7f-8880-d7cbeddfaa5c)

![Image](https://github.com/user-attachments/assets/922c4f2a-c419-442e-9b33-b0f87d974332)

/bounty $25

PR #386, opened 2024-12-15
Show error when manualEdits isn't loaded

RohittCodes:
closes: #370 
/claim #370 
image
on error click ->
image


src/components/CodeEditorHeader.tsx
```diff
@@ -40,6 +40,13 @@ export const CodeEditorHeader = ({
   const [footprintDialogOpen, setFootprintDialogOpen] = useState(false)
   const { toast } = useToast()
 
+  // Check if manual-edits.json has any content
+  const [hasManualEdits, setHasManualEdits] = useState(false)
+
+  useEffect(() => {
+    setHasManualEdits(files["manual-edits.json"].length > 0)
+  }, [files["manual-edits.json"]])
```
seveibar, line 48: this is a DOUBLE anti-pattern 😁 

1. Never use `useEffect` and `useState` when there's nothing async, just use `useMemo`
2. Just use variables if there's no cost to memoization!!!

```ts
const hasManualEdits = files["manual-edits.json"].length > 0
```

src/components/CodeEditorHeader.tsx
```diff
@@ -76,6 +83,68 @@ export const CodeEditorHeader = ({
       })
     }
   }
+
+  const importRegex =
+    /import\s+(?:\*\s+as\s+)?([a-zA-Z_$][\w$]*)\s+from\s+["']\.\/manual-edits\.json["'];?/
```
seveibar, line 88: this code is too specific, it doesn't belong here. Create a utility function called `checkIfManualEditsImported` in a separate file

src/components/CodeEditorHeader.tsx
```diff
@@ -76,6 +83,68 @@ export const CodeEditorHeader = ({
...
+          : match.replace(">", ` manualEdits={${importVariableName}}>`),
+      )
+
+      modifiedContent = modifiedContent.replace(
+        groupSubcircuitRegex,
+        (match) =>
+          match.includes(`manualEdits={${importVariableName}}`)
+            ? match
+            : match.replace(">", ` manualEdits={${importVariableName}}>`),
+      )
```
seveibar, line 117: also isolate this modification to a separate file, we don't want to clutter the `CodeEditorHeader` and have it grow out of control

seveibar (commented):
EXCELLENT, just a minor refactor

Ideally you also have a playwright test for this to make sure others don't break it

seveibar:
You might need to add "click to fix" to that error message because i'm not sure it's obvious you can click it

seveibar (approved):
NOT BAD MY MAN NOT BADDDD

Issue #399, opened 2024-12-19
Persist Snippet Creation Across Login Sessions


Niharika0104:
When a user is not logged in and starts creating a snippet, the snippet is lost after the user clicks the login button.

**Expected Behavior:**
The created snippet should persist across the login process, allowing the user to continue editing or saving it after logging in.

Niharika0104:
@seveibar

PR #400, opened 2024-12-19
will persist snippet data across the login session

Niharika0104:
Fixes - #399

### DIFF
--- a/src/components/CodeAndPreview.tsx
+++ b/src/components/CodeAndPreview.tsx
@@ -25,6 +25,8 @@ interface Props {
 export function CodeAndPreview({ snippet }: Props) {
   const axios = useAxios()
   const isLoggedIn = useGlobalStore((s) => Boolean(s.session))
+  const guestSnippet = useGlobalStore((s) => s.guestSnippet)
+  const setGuestSnippet = useGlobalStore((s) => s.setGuestSnippet)
   const urlParams = useUrlParams()
   const templateFromUrl = useMemo(
     () => getSnippetTemplate(urlParams.template),
@@ -50,14 +52,19 @@ export function CodeAndPreview({ snippet }: Props) {
   const [showPreview, setShowPreview] = useState(true)
   const [lastRunCode, setLastRunCode] = useState(defaultCode ?? "")
   const [fullScreen, setFullScreen] = useState(false)
-
-  const snippetType: "board" | "package" | "model" | "footprint" =
-    snippet?.snippet_type ?? (templateFromUrl.type as any)
+  const [snippetType, setSnippetType] = useState<
+    "board" | "package" | "model" | "footprint" | string | undefined
+  >(snippet?.snippet_type ?? (templateFromUrl.type as any))
 
   useEffect(() => {
     if (snippet?.code) {
       setCode(snippet.code)
       setLastRunCode(snippet.code)
+    } else if (!isLoggedIn && guestSnippet) {
+      setCode(guestSnippet.code)
+      setSnippetType(guestSnippet.snippetType)
+      setGuestSnippet(null)
+      setManualEditsFileContent(guestSnippet.manual_edits_json_content)
     }
   }, [Boolean(snippet)])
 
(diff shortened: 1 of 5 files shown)


src/hooks/use-global-store.ts
```diff
@@ -1,13 +1,22 @@
+import { AnyCircuitElement } from "circuit-json"
 import { create } from "zustand"
 import { persist } from "zustand/middleware"
 
+interface snippetInfo {
+  circuitJson?: AnyCircuitElement[] | null
+  code: string
+  snippetType: "board" | "package" | "model" | "footprint" | string | undefined
```
imrishabh18, line 8: The snippet type shouldn't be `undefined` or `string`, when we are using the const names. It defeats the point of it

src/hooks/use-run-tsx/index.tsx
```diff
@@ -26,7 +26,7 @@ export const useRunTsx = ({
 }: {
   code?: string
   userImports?: Record<string, object>
-  type?: "board" | "footprint" | "package" | "model"
+  type?: "board" | "footprint" | "package" | "model" | string
```
imrishabh18, line 29: Here as well, we shouldn't be using `string`

imrishabh18 (changes requested):

Niharika0104:
@imrishabh18 Okay sure.So we should split tests across multiple files ?

Niharika0104:
@imrishabh18 i'm unable to test these changes locally coz locally we are using fakeApi and not the redirecting the user to the github oauth page
image
the user will click on login to save button and when in prod env it will redirect user to the github oauth page and from there to the authorize page and if the authorize page sees that guest snippet variable exists in the local storage it will redirected to the editor page and I will extract the previous edited info from guest snippet and set it . If that's good then i will start writing a test for this

imrishabh18:
@Niharika0104 Yeah, correct. I think we can merge this without test. What do you think @seveibar?

On the UI, you have converted the `Not logged in, can't save` tag beside the save button to `Login to save` as a button. From the user point of view, that button appears to be a tag, which is pointing me to Login on the header and not click on the tag itself.

Niharika0104:
@imrishabh18 yeah when we click on login to save it will redirect to GitHub oauth page and on entering their credentials it will redirect them to editor page and the code editor and manual edits  json file will be populated with the info we saved in guest snippet variable present in the local storage
Hope this makes sense!!

src/components/CodeAndPreview.tsx
```diff
@@ -50,14 +52,19 @@ export function CodeAndPreview({ snippet }: Props) {
...
-    snippet?.snippet_type ?? (templateFromUrl.type as any)
+  const [snippetType, setSnippetType] = useState<
+    "board" | "package" | "model" | "footprint"
+  >(snippet?.snippet_type ?? (templateFromUrl.type as any))
 
   useEffect(() => {
     if (snippet?.code) {
       setCode(snippet.code)
       setLastRunCode(snippet.code)
+    } else if (!isLoggedIn && guestSnippet) {
```
seveibar, line 63: i think this breaks the loading of templates?

seveibar:
@Niharika0104 I think this should be still testable in playwright if you create a fake-api login route, the url is something like:

- `internal/oauth/github/authorize?next=${nextUrl}/authorize`

Overall i like the implementation, except i think most users will click "login" in the header rather than the badge.

It's a bit of work but if you're down i would recommend pausing this PR, then implementing a PR that just allows logging in inside playwright tests (issue: https://github.com/tscircuit/snippets/issues/409)

Overall it's a good idea to store the guest session inside the global store. But I'm cautious about the loading process- we might want to have a specialized hook or something that runs the appropriate state machine because the `<CodeAndPreview />` file scares me. But I'm not quite sure how to do this refactor yet-

seveibar:
@Niharika0104 i was looking at the logic in a bit more detail, it seems like the playwright test should be able to login due to this login logic:

image

Niharika0104:
Yeah the tests are fine now but when we click on login /sign up on the header it takes us to the home page(/) but when we click on the login to save button it will take us to GitHub oauth and then to the editor page.
I tried testing it in the vercel preview site but GitHub oauth is not working there @seveibar

seveibar:
Yea i cant enable auth in preview

We should use the same code/logic for both logins, there's a sign in hook that should handle it iirc

Niharika0104:
Yeah Im aware of that hook let me try again 
Thanks @seveibar

PR #416, opened 2024-12-24
Revert "fix: Use the template at initialization"

seveibar:
Reverts tscircuit/snippets#413

@imrishabh18 this caused a bug where an Error is shown on every snippet.

I also am not sure that we want to have every snippet require a manual edits file, it should be optional. If it's an empty string/null that means there isn't one set yet

image


seveibar:
@imrishabh18 i'm not sure what bug was being thrown, but it should be fixed in a different way

PR #417, opened 2024-12-24
fix: import manual edits

imrishabh18:
This is the better approach as we are moving to the "pure snippets", I think we shouldn't be throwing error on importing empty file for manual-edits

### DIFF
--- a/src/hooks/use-run-tsx/index.tsx
+++ b/src/hooks/use-run-tsx/index.tsx
@@ -90,6 +90,12 @@ export const useRunTsx = ({
       }
 
       const __tscircuit_require = (name: string) => {
+        if (
+          name === "./manual-edits.json" &&
+          preSuppliedImports["./manual-edits.json"] === ""
+        ) {
+          return preSuppliedImports["./manual-edits.json"]
+        }
         if (!preSuppliedImports[name]) {
           throw new Error(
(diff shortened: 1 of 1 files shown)


seveibar (approved):

seveibar:
@imrishabh18 the playwright test failure might be legitimate here because it involves manual edits

src/hooks/use-run-tsx/index.tsx
```diff
@@ -90,6 +90,12 @@ export const useRunTsx = ({
       }
 
       const __tscircuit_require = (name: string) => {
+        if (
+          name === "./manual-edits.json" &&
+          preSuppliedImports["./manual-edits.json"] === ""
```
seveibar, line 95: we'll probably use null instead of empty string to this in the future

```suggestion
          !preSuppliedImports["./manual-edits.json"]
```

imrishabh18:
@seveibar Checked the manual-edits file, can't see any difference it's just the pcb-viewer lines

Screenshot 2024-12-25 at 17 37 23

I think we are good to merge.

seveibar:
Can you add a network idle check in the playwright test for that file? This is going to keep happening if we dont fox these tests- ansh recently pr’d what that fix may look like

PR #422, opened 2024-12-24
add waitforLoadState to view-snippet.spec.ts test

### LINKED ISSUE #420: Unskip and make sure `view-snippets.spec.ts` passes
This test was super flaky because the snapshot has some timing issue, i think related to @ShiboSoftwareDev 's work (is it possible there's a delay before it switches to the PCB tab on the view snippets page making it flaky?)

When we unskip we should make it not flaky by e.g. waiting on network requests to finish (which is a single playwright function iirc)

/bounty $5

Anshgrover23:
/claim #420 
Fixes #420 

image


seveibar (changes requested):
failing check

Anshgrover23:
@seveibar done.

seveibar (approved):

PR #423, opened 2024-12-24
import snippets from registry to development server

### LINKED ISSUE #174: When running locally, automatically load all snippets from `@tscircuit/featured-snippets` module
The hardest part of this task is the typescript/javascript has to be compiled. This can be accomplished by importing `CodeRunner` from `@tscircuit/prompt-benchmarks`

You'll also have to compile the typescript to dts, this can also be pretty tricky

You should do this inside a vite plugin when the server first starts IF started in development mode (I don't think we want to start it for every playwright test like this)

If it takes a long time to build, we might want to reconsider our approach and load snippets from production or something using `https://registry-api.tscircuit.com/snippets/list_trending`, we could also have a list of snippets to load in using a txt file.

The point is I want some real-world snippets loaded into snippets when developing- it's better for testing things out

ShiboSoftwareDev:
https://github.com/user-attachments/assets/0d5c5192-9087-434a-beb2-de3f2a798ecb

/claim #174
/closes #174

### DIFF
--- a/fake-snippets-api/lib/db/seed.ts
+++ b/fake-snippets-api/lib/db/seed.ts
@@ -1,4 +1,85 @@
-import type { DbClient } from "./db-client"
+import axios from "axios"
+import fs from "fs"
+import path from "path"
+import { DbClient } from "./db-client"
+
+const extractTsciDependencies = (
+  code: string,
+): Array<{ owner: string; name: string }> => {
+  const regex = /@tsci\/([^.]+)\.([^"'\s]+)/g
+  const matches = Array.from(code.matchAll(regex))
+  return matches.map((match) => ({
+    owner: match[1],
+    name: match[2],
+  }))
+}
+
+const registryApi = axios.create({
+  baseURL: "https://registry-api.tscircuit.com",
+  headers: {
+    Accept: "application/json",
+    "Content-Type": "application/json",
+  },
+})
+
+const fetchSnippetFromRegistry = async (owner: string, name: string) => {
+  const response = await registryApi.get(
+    `/snippets/get?owner_name=${owner}&unscoped_name=${name}`,
+  )
+  return response.data.snippet
+}
+
+const loadSnippetWithDependencies = async (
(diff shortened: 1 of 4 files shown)


ShiboSoftwareDev:
cannot find axios? But it worked on my machine

seveibar (commented):
great work, can you make it so that we only do this if we are running the development server and not if we're running tests? worried about increased api test time

fake-snippets-api/lib/db/seed.ts
```diff
@@ -20,6 +27,10 @@ export const seed = (db: DbClient) => {
   db.addAccount({
     github_username: "seveibar",
   })
+
+  // Only load development snippets in development mode
+  loadAutoloadSnippets(db)
```
seveibar, line 32: the if condition should be here

fake-snippets-api/lib/db/seed.ts
```diff
@@ -1,4 +1,11 @@
-import type { DbClient } from "./db-client"
+import { DbClient } from "./db-client"
+
+const loadAutoloadSnippets = async (db: DbClient) => {
+  if (process.env.AUTOLOAD_SNIPPETS === "true") {
```
seveibar, line 4: this function name is misleading because it says `loadAutoloadSnippets` but you're conditionally loading the snippets

i would just move the condition to the call site. But if for whatever reason you wanted this function to contain the condition it would need to be renamed to indicate that it's behavior is condition, `loadAutoloadSnippetsIfConfigured` for example (it's an uncommon way to do things though)

seveibar (commented):
minor change with conditional function antipattern

seveibar (approved):
only slightly hurt you didn't include https://tscircuit.com/seveibar/usb-c-flashlight

ShiboSoftwareDev:
@seveibar I went to add your snippet and it was worth it, I found a bug where I imported duplicate snippets. Fixed now

PR #428, opened 2024-12-26
fix: error message

### LINKED ISSUE #427: Improve error message of the already imported jlcpcb part
The error message is coming from the https://github.com/tscircuit/easyeda-converter, we can modify there and some parts of the message here to convey better message for this case

Image

The error message needs to improve when that component is already imported

Anshgrover23:
/claim #427 
Fixes #427 

#### Related PR:
Fixes here also: https://github.com/tscircuit/easyeda-converter/pull/137

### Screenshots:
- Before
Screenshot from 2024-12-26 19-50-58
- After
Screenshot from 2024-12-26 19-49-59

### DIFF
--- a/src/components/JLCPCBImportDialog.tsx
+++ b/src/components/JLCPCBImportDialog.tsx
@@ -86,13 +86,23 @@ export function JLCPCBImportDialog({
       })
       onOpenChange(false)
       navigate(`/editor?snippet_id=${snippet.snippet_id}`)
-    } catch (error) {
+    } catch (error: any) {
       console.error("Error importing JLCPCB component:", error)
-      toast({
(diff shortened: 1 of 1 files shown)


Anshgrover23:
@imrishabh18 done.

imrishabh18 (commented):
Looking more into this, I think we don't need to change anything in the snippets. Cause this toast is not visible, as we already have the dailog box for the already imported snippet, and the error message changed in the easyeda will be propagated here.

cc: @seveibar

Anshgrover23:
@imrishabh18 yes i also think so.

seveibar (changes requested):
this is a server-side issue

PR #443, opened 2024-12-29
add quantity column in bills

### LINKED ISSUE #415: Make a quantity column to prevent superlong BOM
This is what we have now.
Image

I have this in mind. It will be clean, and we will keep track of the quantity of every part.
Image

Notice that they share the same row when there is more than one name with the same part number.

cc @seveibar

techmannih:
fix #415
/claim #415


seveibar:
Can you make a playwright test that snapshots the BOM tab content on a representative page?

seveibar:
Also this fix is in the wrong repo, we're migrating to runframe, so youll need to make the change there, you can close this

Issue #473, opened 2025-01-02
Allow downloading of assembly image


seveibar:
/bounty $2

PR #475, opened 2025-01-03
add download Assembly svg button

techmannih:
fic #473 
/claim #473


seveibar:
can you demonstrate this works either with a video or playwright test?

seveibar (approved):
nice work!

seveibar:
yes we need to integrate runframe first

PR #485, opened 2025-01-04
footprint dialog enhancements

ShiboSoftwareDev:
- [x] don't display passive components as options
- [x] you can now scroll on mobile using combobox
- [x] readonly params will be displayed but editing them has no effect
- [x] special handling for bga
- [x] now all footprints and options work perfectly

/claim https://github.com/tscircuit/footprinter/issues/88


seveibar (approved):

seveibar:
@ShiboSoftwareDev yes your suggestion seems ok!!

The scope of this feature will include schPortArrangement in the future as well? That would be nice but could also be a separate thing?

ShiboSoftwareDev:
@seveibar let's get this working 100% before we do the schematic

PR #487, opened 2025-01-04
Revert "fix card misalignment"

seveibar:
Reverts tscircuit/snippets#478

@techmannih this broke a bunch of pages, make sure to not change the ShadCN components, because it impacts every page


seveibar:
@techmannih 
image

PR #492, opened 2025-01-05
fixes the selector

Anshgrover23:


seveibar (changes requested):
check your diffs, you committed some crazy file

Anshgrover23:
@seveibar this is due to bun i i have removed that?

seveibar (approved):

Issue #388, opened 2024-12-17
manual-edits.json file is not being detected or recognized by the editor


rohittcodes:
![Image](https://github.com/user-attachments/assets/01d88b56-e1a4-41d9-ae76-386e85bc129d)

Bhavyajain21:
Do we have this file in the code base? @imrishabh18 / @RohittCodes

imrishabh18:
@Bhavyajain21 Yes, here is the template which is used for the `manual-edits.json` [src/lib/templates/manual-edits-template.ts](https://github.com/tscircuit/snippets/blob/e6f7c79d053095bf9b16de812194ec18de0d2094/src/lib/templates/manual-edits-template.ts)

PR #500, opened 2025-01-07
Add type declaration for manual-edits.json in ATA

oldbear26:
/claim #388

### DIFF
--- a/src/components/CodeEditor.tsx
+++ b/src/components/CodeEditor.tsx
@@ -154,6 +154,19 @@ export const CodeEditor = ({
         return fetch(input, init)
       },
       delegate: {
+        started: () => {
+          const manualEditsTypeDeclaration = `
+				  declare module "*.json" {
+				  const value: {
+					  pcb_placements?: any[],
+					  edit_events?: any[],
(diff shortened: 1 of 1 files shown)


Anshgrover23 (changes requested):
@oldbear26 add any screenshot or video that this is working? also, why are you hardcoding the values.

seveibar (approved):

src/components/CodeEditor.tsx
```diff
@@ -154,6 +154,19 @@ export const CodeEditor = ({
         return fetch(input, init)
       },
       delegate: {
+        started: () => {
+          const manualEditsTypeDeclaration = `
+				  declare module "*.json" {
```
seveibar, line 159:
```suggestion
				  declare module "manual-edits.json" {
```

seveibar: another option, this is ok for now though

seveibar:
@Anshgrover23 this file has slightly special handling so @oldbear26 is kind of correct for the injection- we handle it even if it's not defined etc.

PR #508, opened 2025-01-08
fixed autoload snippets for mac

ShiboSoftwareDev:


seveibar:
this is the fix, i guess maybe windows has different handling: https://github.com/tscircuit/snippets/pull/512

PR #523, opened 2025-01-10
 bundles the manual-edits.json file into snippet.compiled_js

Abse2001:
/claim https://github.com/tscircuit/eval-webworker/issues/66
in this screenshot I added manual edits to the compiled code and removed any import now the manual edits file content are inside the code and does not need to be moved imported but still you need to use the manual edits prop to use it,
I want your opinion on how to handle the manual edits prop, for eg we can just define var let manualEdits and pass it down to the prop and it will be filled with the manual edit conten on compile

https://github.com/user-attachments/assets/1b581a49-f0b8-47df-ad1a-725d9c80721f

image
image
image
CC @seveibar

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ test-results
 yalc.lock
 
 public/assets
+public/sitemap.xml


src/hooks/use-compiled-tsx.ts
```diff
@@ -3,14 +3,27 @@ import * as Babel from "@babel/standalone"
...
+        .trim()
+        .replace(/manualEdits={[^}]+}/, "manualEdits={__MANUAL_EDITS}")
+    }
+
     return {
       success: true,
       compiledTsx:
-        Babel.transform(code, {
+        Babel.transform(codeWithManualEdits, {
           presets: ["react", "typescript"],
```
seveibar, line 27: is it possible to use a bundle step here, and use `userImports` instead of taking the manualEditsFileContent? We're going to want to bundle other things in the future

seveibar: @Abse2001 , we may need to incorporate something like rollup....

image

seveibar: @Abse2001 it is possible that if we support web standards properly, we could support browser-native dynamic `import` and not need to bundle anything. However, this is a bit complex and I'll need to take some time to understand the browser specifications for dynamically importing

seveibar: I'm investigating this extensively right now and want to help more... here's the doc on how to run rollup in the browser https://rollupjs.org/faqs/#how-do-i-run-rollup-itself-in-a-browser

I'm trying to figure out if we can somehow build to a format that's a bit easier to work with, but it's very tricky to say the least...

PR #543, opened 2025-01-12
improve seo by adding robot file and generating sitemap for crawlers

ShiboSoftwareDev:
attempt to fix this:
image


public/robots.txt
```diff
@@ -0,0 +1,14 @@
...
+User-agent: *
+Allow: /
+
+# Sitemaps
+Sitemap: https://tscircuit.com/sitemap.xml
+
+# Disallow specific paths
+Disallow: /api/
+Disallow: /dev-login
+Disallow: /authorize
```
seveibar, line 11: haha we don't really have any of these routes, maybe comment it out for now? Just to avoid confusion

public/sitemap.xml
```diff
@@ -0,0 +1,67 @@
...
+  <url>
+    <loc>https://tscircuit.com/legal/terms-of-service</loc>
+    <changefreq>yearly</changefreq>
+    <priority>0.3</priority>
+  </url>
+  <url>
+    <loc>https://tscircuit.com/legal/privacy-policy</loc>
+    <changefreq>yearly</changefreq>
+    <priority>0.3</priority>
+  </url>
```
seveibar, line 66: ok this is pretty good but i would say that we should try to generate this using `/snippets/list_trending` and include all the trending snippets etc.

This is what many sites do to make sure that popular snippets are indexed.

You can create an issue from this comment to do in a future PR, but I would consider making this file generated with a script (there are sitemap builder libraries on npm i think) just to make sure people understand we shouldn't have this be static- it's a build artifact

seveibar (approved):

public/sitemap.xml
ShiboSoftwareDev: I agree

ShiboSoftwareDev:
now adds latest snippets
image

seveibar (approved):
NICE /tip $10

PR #556, opened 2025-01-13
some accessibility fields and eliminated layout shift on landing page

ShiboSoftwareDev:
100 accessibility score, and no more layout shift
image


seveibar (approved):
NICE!!!

PR #557, opened 2025-01-13
add useMychip hook

### LINKED ISSUE #555: Circuit module template should include `useMyChip` hook with `createUseComponent` and a pinLabels definition
Put the pinLabels at the bottom instead of the top, it's distracting at the top

https://tscircuit.com/editor?snippet_id=51cb0ad7-3dea-4727-b035-b15b748bee23

techmannih:
fixes #555 
/claim #555

### DIFF
--- a/src/lib/templates/blank-package-template.ts
+++ b/src/lib/templates/blank-package-template.ts
@@ -4,37 +4,17 @@ export const blankPackageTemplate = {
 import { createUseComponent } from "@tscircuit/core"
 
 export const MyChip = (props: { name: string }) => (
-  <chip
-    {...props}
-    pinLabels={pinLabels}
-    footprint="soic8"
-  />
+  <chip {...props} pinLabels={pinLabels} footprint="soic8" />
 )
 const pinLabels = {
-  "pin1": [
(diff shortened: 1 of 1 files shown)


src/lib/templates/blank-package-template.ts
```diff
@@ -1,11 +1,56 @@
...
+  ],
+  "pin7": [
+    "pin7",
+    "AGND"
+  ],
+  "pin8": [
+    "pin8",
+    "GND3"
+  ]
+} as const
```
seveibar, line 52: the redundant "pin1": ["pin1", ...] isn't needed anymore.

seveibar (changes requested):
make sure to make this code formatted as it would be if you ran format in the editor, looks messy

src/lib/templates/blank-package-template.ts
seveibar: just make in an empty array, and please format this

seveibar (changes requested):
the redundant pin isn't needed....

seveibar (approved):

Issue #533, opened 2025-01-11
implement a simple fancy loading page, for all navigation


ShiboSoftwareDev:
![Image](https://github.com/user-attachments/assets/ed29f552-4e8d-4c05-b919-45545f359212)

seveibar:
/bounty $10

YES skeletons are ideal- let's have a general purpose skeleton

PR #566, opened 2025-01-14
added light weight skeleton loader for navigation

bhuvan-crypto:
/claim #533

https://github.com/user-attachments/assets/580e1947-151a-4ce8-821d-85bfa00c4551


seveibar:
You didn't understand the issue, you can't use the bundled javascript for this skeleton.

bhuvan-crypto:
> You didn't understand the issue, you can't use the bundled javascript for this skeleton.

My bad so what i understood now that the loader to be placed in index.html and from there it should be used

PR #574, opened 2025-01-15
Improve error message for 3D model download when 3D viewer is not open

### LINKED ISSUE #563: If 3d download fails because 3d viewer isn't open, show a more appropriate error
If there isn't a 3d viewer ref, we should tell the user to open the 3d viewer

/bounty $2

techmannih:
fixes #563 
/claim #563

### DIFF
--- a/src/lib/download-fns/download-gltf.ts
+++ b/src/lib/download-fns/download-gltf.ts
@@ -12,7 +12,9 @@ export const downloadGltf = async (
     ?.current as THREE.Object3D
 
   if (!threeJsObject) {
-    throw new Error("To download the 3D model, please open the 3D view first")
+    throw new Error(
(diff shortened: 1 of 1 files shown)


src/lib/download-fns/download-gltf.ts
```diff
@@ -12,9 +12,7 @@ export const downloadGltf = async (
     ?.current as THREE.Object3D
 
   if (!threeJsObject) {
-    throw new Error(
-      "No 3D object found, run the snippet before downloading the 3d model",
-    )
+    throw new Error("To download the 3D model, please open the 3D view first")
```
seveibar, line 15:
```suggestion
    throw new Error("To download the 3D model, please open the 3D view first and run the snippet")
```

seveibar (approved):

PR #627, opened 2025-01-31
add fake-api for the packages

imrishabh18:


seveibar (approved):
@imrishabh18 this is fine but can you follow up and remove the `snippets` inside the database? As you know on production packages and snippets are not represented with separate tables, a snippet is just a package. I'm worried that this will create a lot of issues in our fake implementation

PR #630, opened 2025-02-02
add 3d svg support in fake-snippets-api

Anshgrover23:


imrishabh18 (commented):
Yes, the approach is fine. But I would say wait for this PR, as I am currently fixing the `3d-viewer` camera angle and maybe I will simplify the prop being passed to `convertCircuitJsonTo3dSvg`. It would be better if we merge this after that update

PR #636, opened 2025-02-03
feat(fake-snippets-api): missing package_release/get logic

ArnavK-09:


imrishabh18:
It is not missing logic, our `registy-api` doesn't use the `package_name_with_version` so we are just faking it here the same way.

Probably we need to update the `package_name_with_version` in the registry-api as well.

cc: @seveibar

ArnavK-09:
> It is not missing logic, our `registy-api` doesn't use the `package_name_with_version` so we are just faking it here the same way.
> 
> Probably we need to update the `package_name_with_version` in the registry-api as well.
> 
> cc: @seveibar

uhhh, but in production api bump thing is working with same prop

https://github.com/tscircuit/cli/pull/44/files#diff-e9888b99d1bcfe8382103cbee5f41ef1caa69947005a8673481585756b43c89bR125

PR #637, opened 2025-02-04
chore: update packages (fix format issue)

imrishabh18:
Screenshot 2025-02-04 at 21 42 32


imrishabh18:
@seveibar Is this something with vercel? cause when you see the preview deployment of this PR. It works fine! but when merged to main it broke the snippet view

PR #663, opened 2025-02-10
Remove console.logs, allow viewing core version, update props

seveibar:


src/hooks/use-run-tsx/construct-circuit.tsx
```diff
@@ -27,6 +27,7 @@ export const constructCircuit = ({
     const board = circuit.root?.children[0]
     // const board = circuit.selectOne("board")
     if (board) {
+      console.log("board props", board.props)
```
seveibar, line 30:
```suggestion
```

PR #681, opened 2025-02-17
feat: show warning if the snippet is not saved

### LINKED ISSUE #680: Snippets code changes are not saved until you `Run` the snippet and then save
Snippet Toast should give the warning or error that the snippet is not saved until you run the snippet. Currently, it gives successful message which is wrong it is not saved!

https://github.com/user-attachments/assets/31d8acd3-c839-4a65-834e-46634ad61cde

krushnarout:
fixes #680 
/claim #680 

https://github.com/user-attachments/assets/6303df11-f92a-4faf-bbc9-f44bba24ce79

### DIFF
--- a/src/components/CodeAndPreview.tsx
+++ b/src/components/CodeAndPreview.tsx
@@ -166,6 +166,15 @@ export function CodeAndPreview({ snippet }: Props) {
   const [lastSavedAt, setLastSavedAt] = useState(Date.now())
 
   const handleSave = async () => {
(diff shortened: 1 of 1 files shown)


imrishabh18 (approved):

seveibar (approved):
@imrishabh18 fwiw this workflow sucks, you shouldn't need to run to save- we should just run on cloud and never use the user's circuit json when they save, this also means that circuit json will be more deterministic (especially for people saving with old CLI versions)

PR #700, opened 2025-03-01
update the snippet create fake

imrishabh18:


seveibar (approved):

bun-tests/fake-snippets-api/fixtures/get-test-server.ts
```diff
@@ -37,7 +37,7 @@ export const getTestServer = async (): Promise<TestFixture> => {
     if (server && typeof server.stop === "function") {
       await server.stop()
     }
-    // Here you might want to add logic to drop the test database
+    await db.dropDatabase()
```
seveibar, line 40: oh this isn't necessary, a new db is created on every test

fake-snippets-api/lib/db/db-client.ts
```diff
@@ -277,32 +275,29 @@ const initializer = combine(databaseSchema.parse({}), (set, get) => ({
     }))
     return newLoginPage
   },
-  getLoginPage: (login_page_id: string): LoginPage | undefined => {
+  getLoginPage: (loginPageId: string): LoginPage | undefined => {
```
seveibar, line 278: so b/c of variable transparency this change isn't really needed i think

seveibar: it's fine tho

Issue #694, opened 2025-02-26
`Save` doesn't work after running the snippet in webworker


imrishabh18:
After running the snippet in the webworker and clicking on `Save` gives error and doesn't saves.

<img width="1781" alt="Image" src="https://github.com/user-attachments/assets/10c43250-a04a-4c1a-97f8-ff1ad597b2ea" />

seveibar:
@imrishabh18 i think this is only on `/editor`, we are deprecating that page in favor of `/p/editor`

PR #702, opened 2025-03-02
Remove legacy tsx runner, replace with RunFrame, everything uses WebWorkers

seveibar:


seveibar:
fixes https://github.com/tscircuit/tscircuit.com/issues/694

src/components/CodeAndPreview.tsx
```diff
@@ -112,7 +88,7 @@ export function CodeAndPreview({ snippet }: Props) {
         snippet_id: snippet.snippet_id,
         code: code,
         dts: dts,
-        compiled_js: compiledJs,
+        // compiled_js: compiledJs,
```
seveibar, line 91: @imrishabh18 note this is one of the biggest changes, we are no longer sending compiled_js to the server. HOWEVER, this was only needed for legacy tscircuit runners, there is now a server bundler/caching system that should eliminate the need for clients (web and cli) to compile JS or compile typescript (the dts)

Still we should be mindful that snippets saved with this will be slightly different

Issue #725, opened 2025-03-11
Fix the error message and support snippet name with spaces


imrishabh18:
- [ ] Modify the error message that the spaces are not available in snippet name
- [ ] Support snippet name with spaces in fake

<img width="1345" alt="Image" src="https://github.com/user-attachments/assets/85dbee4e-41e4-4930-8458-25a41167bc16" />

siva222003:
@imrishabh18 could you please review the PR?

PR #736, opened 2025-03-21
feat: prevent spaces in snippet name and show error message

siva222003:
## Fix: #725

- Added validation to prevent spaces in snippet names and avoid unnecessary API calls.  
- Displayed an error message when a user tries to enter a name with spaces.  

[Screencast from 2025-03-21 18-31-21.webm](https://github.com/user-attachments/assets/1c1db63b-8771-4469-9d30-e0e97859b9f1)  

Please let me know if any changes are required.

### DIFF
--- a/src/components/dialogs/rename-snippet-dialog.tsx
+++ b/src/components/dialogs/rename-snippet-dialog.tsx
@@ -6,6 +6,7 @@ import { useMutation, useQueryClient } from "react-query"
 import { createUseDialog } from "./create-use-dialog"
 import { useAxios } from "@/hooks/use-axios"
 import { useToast } from "@/hooks/use-toast"
+import { useIsUsingFakeApi } from "@/hooks/use-is-using-fake-api"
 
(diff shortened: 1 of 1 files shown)


imrishabh18 (changes requested):
I don't think this is the right implementation. Better would be the backend to handle this and send the error message.

siva222003:
> I don't think this is the right implementation. Better would be the backend to handle this and send the error message.

@imrishabh18 Yeah, that would be better, but since the issue was raised here, I thought of resolving it from the frontend itself. This way, there's no need to send an API request.

imrishabh18:
You can check the `fake-snippet` directory, it has the backend routes where you can add this. If all looks good, I can then implement the same logic in the prod backend

siva222003:
> You can check the `fake-snippet` directory, it has the backend routes where you can add this. If all looks good, I can then implement the same logic in the prod backend

Alright, will implement that.

Issue #678, opened 2025-02-16
Auto format code on import


imrishabh18:
After importing the code from the jlcpcb auto format the code

siva222003:
Hey @imrishabh18, is this issue still open? My PR was closed due to inactivity. If it’s still open, I’d be happy to raise it again.

imrishabh18:
@siva222003 I think so, probably you can verify it in the local by pulling the latest `main`. Raise a PR, I will take a look and merge it.

siva222003:
@imrishabh18 I have already tested it, and it is still not resolved. Raising a PR, please review it.

siva222003:
Hey @imrishabh18 Could you please review the PR again?

PR #737, opened 2025-03-21
feat(editor): auto-format code before opening editor

siva222003:
##Fix: #678 

- Refactored the code formatting logic into a reusable utility function.
- Now, the code is auto-formatted before opening the editor.
- Added auto-formatting when importing from JLCPCB.

[Screencast from 2025-02-23 19-51-24.webm](https://github.com/user-attachments/assets/517672d3-365b-4023-a4cc-e677cf52e399)

Let me know if any changes are needed!

### DIFF
--- a/src/lib/utils/formatCode.ts
+++ b/src/lib/utils/formatCode.ts
@@ -12,18 +12,26 @@ export const formatCode = ({
   if (!window.prettier || !window.prettierPlugins || !currentContent)
     return null
 
-  try {
-    if (currentFile.endsWith(".json")) {
+  if (currentFile.endsWith(".json")) {
+    try {
       return JSON.stringify(JSON.parse(currentContent), null, 2)
+    } catch (error) {
+      console.error("JSON Formatting Error: Invalid JSON format", error)
+      return null
     }
+  }
 
+  try {
     return window.prettier.format(currentContent, {
       semi: false,
       parser: "typescript",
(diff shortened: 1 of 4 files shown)


src/lib/utils/formatCode.ts
```diff
@@ -0,0 +1,29 @@
...
+  try {
+    if (currentFile.endsWith(".json")) {
+      return JSON.stringify(JSON.parse(currentContent), null, 2)
+    }
+
+    return window.prettier.format(currentContent, {
+      semi: false,
+      parser: "typescript",
+      plugins: window.prettierPlugins,
+    })
```
imrishabh18, line 24: We are having two conditions inside the try-catch block, we can separate out this into separately and throw error message specific to the file extension

imrishabh18 (approved):
See the comment, I think we are good to merge after that

src/lib/utils/formatCode.ts
siva222003: Hey @imrishabh18, I have refactored the try catch to throw the seperate file extension errors. Do you want me to show any toast when formatting is failed or this is enough

siva222003:
@imrishabh18 Could you please review this?

imrishabh18:
LGTM! Let @seveibar take a look

seveibar (changes requested):
We want to format jlcpcb imported components, not every time code is opened

Issue #740, opened 2025-03-24
Bad alignment, generally bad looking cards on the user profile in packages page


seveibar:
![Image](https://github.com/user-attachments/assets/48894fe3-c16f-46e8-8b91-19db8b60abc8)

Please show a thumbnail of the PCB in the preview

Make sure all text is aligned

Some inspiration:

## Replit

![Image](https://github.com/user-attachments/assets/295ae455-ae3c-4b53-80df-797969dae793)

## val town

![Image](https://github.com/user-attachments/assets/717c42ee-384e-4b05-9e8c-18d064625a43)

/bounty $8

PR #741, opened 2025-03-24
fix(ui): improve profile cards design and fix layout issues

onyedikachi-david:
Fixes: #740
/claim #740

image


Anshgrover23 (changes requested):
@onyedikachi-david Thanks for contributing to TSCircuit. Can u just check why this profile-one playwright test failing i think it is due to your changes at current only one playwright should be failed for now?
image

PR #758, opened 2025-03-27
Update fake `/order/create` to handle simulate scenarios

imrishabh18:


fake-snippets-api/routes/api/orders/create.ts
```diff
@@ -1,36 +1,63 @@
...
+import { orderSchema, errorSchema, Order } from "fake-snippets-api/lib/db/schema"
+
+const simulate_scenarios = [
+  // TODO: Add more scenarios
+  "gerber-generation-failed",
+  "gerber-upload-failed",
+  "bom-generation-failed",
+  "bom-upload-failed",
+  "pnp-generation-failed",
+  "pnp-upload-failed",
```
seveibar, line 12: we should use underscore, this is my mistake, i'm trying to be consistent on the underscore for enum convention but i forget all the time

the reason why: when you make an sdk you can't code enums if there are dashes

seveibar (approved):
pretty good!

note the enum thing

PR #761, opened 2025-03-27
Modify logic for `/order/create` and support polling for `/order/get`

imrishabh18:


fake-snippets-api/lib/db/schema.ts
```diff
@@ -74,7 +74,7 @@ export const accountSchema = z.object({
 export type Account = z.infer<typeof accountSchema>
 
 export const orderSchema = z.object({
-  order_id: z.string(),
+  order_id: z.string().nullable(),
```
seveibar, line 77: Its a bit weird for this to be nullable (why?)

fake-snippets-api/routes/api/orders/create.ts
```diff
@@ -198,11 +197,6 @@ export default withRouteSpec({
     newOrder.has_error = true
   }
 
-  if (newOrder.has_error) {
-    return ctx.json({
-      order: newOrder,
-    })
-  }
```
seveibar, line 205: If the order fails to create, we just return { error }

seveibar: For simulated errors, you have to poll for them just like the real api, they dont happen instantly

seveibar: Ok i see whats going on. I think we still add the order to the fb if there’s an error, the real behavior is the order gets added, then asynchronously in the job queue it gets the error attached

fake-snippets-api/routes/api/orders/create.ts
```diff
@@ -198,11 +197,6 @@ export default withRouteSpec({
...
   const order = ctx.db.addOrder(newOrder)
```
seveibar, line 200: Also /orders/create should never return an order with an error, the error is always found on the mext poll, because this is the behavior of the real api

seveibar: If for some reason we cant create the order synchronously (eg bad request params) we return { error } and dont create an order. But this doesnt meed the simulate scenario

fake-snippets-api/routes/api/orders/get.ts
```diff
@@ -26,6 +49,90 @@ export default withRouteSpec({
...
+        error_code: "INITIAL_COST_CALCULATION_FAILED",
+        message: "Initial cost calculation failed",
+      },
+    }
+
+    const error = errorMap[_simulate_scenario]
+    if (error) {
+      order.error = errorSchema.parse(error)
+      order.has_error = true
+      ctx.db.updateOrder(order_id, { error: order.error, has_error: true })
```
seveibar, line 132: i think we still need binary flags on the order in order to display the checklist. You could put them inside an object, e.g. `order.jlcpcb_order_state = { are_gerbers_uploaded:false, ... }`

imrishabh18: If you look at the response being sent by this endpoint, the orderState has this data.

imrishabh18:
@seveibar Is this api okay to go ahead with?

PR #769, opened 2025-03-31
modify SEO tags in middleware

seveibar:


seveibar:
CC @Ayushjhawar8 this PR should do the trick. The alternative (better) approach would be to get SSR woorking with vite, but I wasn't able to (see #768), but SSR is much better because other page contents can load which are much better for crawlers

This fixes the major issue where canonical urls were wrong, so I'm going to merge because the canonical urls are killing our SEO so badly

Issue #767, opened 2025-03-30
Improve broken image thumbnails


seveibar:
/bounty $15

Also happens in search box.

![Image](https://github.com/user-attachments/assets/f757f9f6-ce1a-42b3-a668-73332ae9e09e)

Needs to look good!!!

PR #776, opened 2025-03-31
Added fallback for broken images

MustafaMulla29:
/claim #767


MustafaMulla29:
@imrishabh18 added just a empty board svg as a fallback. Let me if i have to change to something different

MustafaMulla29:
image

public/images/fallback-image.svg
Anshgrover23, line 1: add this file in assests folder in public @MustafaMulla29 !!

Anshgrover23 (changes requested):

Anshgrover23 (changes requested):
Also, how about if we add playwright for this thing ?
CC: @seveibar

imrishabh18 (commented):
The first one would be better (blank board), no need for the holes on the board as a fallback circuit image

MustafaMulla29:
> @MustafaMulla29 this issue is solved or i can proceed?

Under review, will let you know

seveibar (approved):

Anshgrover23 (approved):

PR #791, opened 2025-04-02
autoload fetching ported from /snippets to /packages

### LINKED ISSUE #786: Port autoloaded snippets to load using the packages api (in order to get `package.ai_usage` and `package.ai_usage_instructions`)
Modify fake-snippets-api/lib/db/autoload-dev-snippets.ts to use the `/packages/*` api instead of the snippets api for both fetching and loading into the local database.

Please prove in some way, e.g. via video, that autoloading is still functional (there are not autoloading tests)

/bounty $15

imrishabh18:
https://github.com/user-attachments/assets/34aa5bde-2b66-430a-a99c-a29561412038

/close #786


fake-snippets-api/lib/db/autoload-packages.json
```diff
@@ -1,13 +1,5 @@
 {
-  "snippets": [
-    {
-      "owner": "Abse2001",
-      "name": "Arduino-Nano-Servo-Breakout"
-    },
-    {
-      "owner": "ShiboSoftwareDev",
-      "name": "Wifi-Camera-Module"
-    },
```
seveibar, line 10: i like these packages!

seveibar (approved):

fake-snippets-api/lib/db/autoload-packages.json
imrishabh18: Will add those back, they were kind of failing to fetch majority of time. So removed it

seveibar: hmmm weird, maybe the package files are too big? I think these have `dist/circuit.json` so I wonder if that could be related

PR #795, opened 2025-04-02
Hide lockfiles and config files.

imrishabh18:
Screenshot 2025-04-02 at 23 33 09


seveibar (approved):
I dont think you generated dist (as mentioned in the title), thats something that happens on the backend

Issue #803, opened 2025-04-04
Don't show schematic/pcb/3d view buttons if they aren't available for a package


seveibar:
![Image](https://github.com/user-attachments/assets/1dc79c7d-5f0f-40c0-b330-2bd7687432ee)

/bounty $5

PR #804, opened 2025-04-05
Don't show schematic/pcb/3d view buttons if they aren't available for a package 

MustafaMulla29:
/claim #803

image

image

### DIFF
--- a/src/components/ViewPackagePage/components/mobile-sidebar.tsx
+++ b/src/components/ViewPackagePage/components/mobile-sidebar.tsx
@@ -31,7 +31,6 @@ export default function MobileSidebar({
   const topics = packageInfo?.is_package ? ["Package"] : ["Board"]
   const { availableViews } = usePreviewImages({
     packageName: packageInfo?.name,
-    onViewChange,
   })
 
   const handleViewClick = (viewId: string) => {
--- a/src/components/ViewPackagePage/components/mobile-sidebar.tsx
+++ b/src/components/ViewPackagePage/components/mobile-sidebar.tsx
@@ -29,6 +29,7 @@ export default function MobileSidebar({
   onViewChange,
 }: MobileSidebarProps) {
   const topics = packageInfo?.is_package ? ["Package"] : ["Board"]
+
   const { availableViews } = usePreviewImages({
     packageName: packageInfo?.name,
   })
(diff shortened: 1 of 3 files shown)


imrishabh18 (changes requested):
I don't think this is the right fix, cause the 3d.png takes more time compared to the others. Currently, we are not showing that and when the response does come for the 3d.png it abruptly shows it. Maybe a skeleton or something could be better as a default

MustafaMulla29:
@imrishabh18 

https://github.com/user-attachments/assets/8812deb0-7997-4183-8175-128fe318103d

is this how it should be?(don't know why there is a delay in the skeleton here, eveything's fine but idk what's the cause)

seveibar (changes requested):
Isolate to a hook, make sure you fix both mobile and desktop (i think you only fixed one)

hook could be usePreviewImages and could return an array of image objects with status attached. Or could return the views

src/hooks/use-preview-images.ts
```diff
@@ -0,0 +1,78 @@
...
+      {
+        id: "schematic",
+        label: "Schematic View",
+        imageUrl: packageName
+          ? `https://registry-api.tscircuit.com/snippets/images/${packageName}/schematic.png`
+          : undefined,
+      },
+    ],
+    [packageName],
+  )
```
seveibar, line 45: Dont use useMemo for simple object constructions, it’s actually less performant

src/hooks/use-preview-images.ts
```diff
@@ -0,0 +1,78 @@
...
+      ...view,
+      status: imageStatus[view.id],
+    }))
+    .filter((view) => view.status !== "error")
+
+  return {
+    availableViews,
+    handleImageLoad,
+    handleImageError,
+    handleViewClick,
```
seveibar, line 76: Dont expose image handling to the hook user, the hook user just needs to get the available views

seveibar (commented):
Looks good but see comments

src/components/ViewPackagePage/components/mobile-sidebar.tsx
```diff
@@ -1,6 +1,8 @@
+"use client"
```
seveibar, line 1: This isnt nextjs, these pragmas arent used

seveibar:
```suggestion
```

seveibar (approved):

PR #805, opened 2025-04-05
Fake for order-quote (WIP)

imrishabh18:
@seveibar We can convert the circuit_json to bom inside the create endpoint and then that will be picked up by the job to get the price of that parts and then update the orderQuote. Or we don't store the bom in the orderQuotes and just the components (part numbers) whic h is good enough for the job to do the calculation and update the orderQuote?

### DIFF
--- a/fake-snippets-api/lib/db/schema.ts
+++ b/fake-snippets-api/lib/db/schema.ts
@@ -101,14 +101,11 @@ export const orderFileSchema = z.object({
 export type OrderFile = z.infer<typeof orderFileSchema>
 
 export const orderQuoteComponentSchema = z.object({
-  component_id: z.string(),
-  name: z.string(),
   manufacturer_part_number: z.string().nullable(),
-  jlcpcb_part_number: z.string().nullable(),
-  quantity: z.number(),
-  unit_price: z.number(),
-  total_price: z.number(),
-  type: z.string(), // e.g., "resistor", "capacitor", etc.
+  supplier_part_number: z.string().nullable(),
+  quantity: z.number().default(0),
+  unit_price: z.number().default(0),
+  total_price: z.number().default(0),
   available: z.boolean().default(true),
 })
 export type OrderQuoteComponent = z.infer<typeof orderQuoteComponentSchema>
@@ -124,10 +121,9 @@ export const orderQuoteSchema = z.object({
   created_at: z.string(),
   updated_at: z.string(),
   completed_at: z.string().nullable(),
-  bom_rows: z.string(),
-  components: z.array(orderQuoteComponentSchema).default([]),
+  quoted_components: z.array(orderQuoteComponentSchema).default([]),
   pcb_price: z.number().default(0),
-  components_price: z.number().default(0),
+  total_components_price: z.number().default(0),
   shipping_price: z.number().default(0),
   total_price: z.number().default(0),
 })
--- a/fake-snippets-api/lib/db/schema.ts
+++ b/fake-snippets-api/lib/db/schema.ts
@@ -100,6 +100,39 @@ export const orderFileSchema = z.object({
 })
 export type OrderFile = z.infer<typeof orderFileSchema>
 
+export const orderQuoteComponentSchema = z.object({
+  component_id: z.string(),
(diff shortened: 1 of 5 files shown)


fake-snippets-api/lib/db/schema.ts
```diff
@@ -100,6 +100,39 @@ export const orderFileSchema = z.object({
...
+  account_id: z.string().nullable(),
+  package_release_id: z.string().nullable(),
+  is_completed: z.boolean().default(false),
+  is_processing: z.boolean().default(true),
+  error: errorSchema.nullable(),
+  has_error: z.boolean().default(false),
+  created_at: z.string(),
+  updated_at: z.string(),
+  completed_at: z.string().nullable(),
+  bom_rows: z.string(),
```
seveibar, line 127: I dont know if we need to store a string here

fake-snippets-api/lib/db/schema.ts
```diff
@@ -100,6 +100,39 @@ export const orderFileSchema = z.object({
...
+  components: z.array(orderQuoteComponentSchema).default([]),
```
seveibar, line 128: lets use a more universally unique key here:

quoted_components

fake-snippets-api/lib/db/schema.ts
imrishabh18: So, only the part numbers from the bom values?

fake-snippets-api/lib/db/schema.ts
```diff
@@ -100,6 +100,39 @@ export const orderFileSchema = z.object({
...
+export const orderQuoteComponentSchema = z.object({
+  component_id: z.string(),
+  name: z.string(),
+  manufacturer_part_number: z.string().nullable(),
+  jlcpcb_part_number: z.string().nullable(),
+  quantity: z.number(),
+  unit_price: z.number(),
+  total_price: z.number(),
+  type: z.string(), // e.g., "resistor", "capacitor", etc.
+  available: z.boolean().default(true),
```
seveibar, line 112: Instead of jlcpcb_part_number, can you use supplier_part_numbers

no need for the type property

fake-snippets-api/lib/db/schema.ts
```diff
@@ -100,6 +100,39 @@ export const orderFileSchema = z.object({
 })
 export type OrderFile = z.infer<typeof orderFileSchema>
 
+export const orderQuoteComponentSchema = z.object({
+  component_id: z.string(),
+  name: z.string(),
```
seveibar, line 105: You can’t have both a name and a quantity, if the name is the same there is at most quantity 1

btw it might be easier to just use the bom row format, which has each name on a separate line. I dont know where that is spec’d though- maybe in corcuit-json-to-bom

```suggestion
```

fake-snippets-api/lib/db/schema.ts
```diff
@@ -100,6 +100,39 @@ export const orderFileSchema = z.object({
 })
 export type OrderFile = z.infer<typeof orderFileSchema>
 
+export const orderQuoteComponentSchema = z.object({
+  component_id: z.string(),
```
seveibar, line 104:
```suggestion
  quoted_component_id: z.string(),
```

seveibar: Honestly im not really sure we need this key, maybe remove?

seveibar:
The BOM will most likely be computed inside the job, not on create. So the quoted_components can be nullable initially

imrishabh18:
> The BOM will most likely be computed inside the job, not on create. So the quoted_components can be nullable initially

I am using the same request body as the `/create/order` as mentioned by you, so if BOM will be calculated inside the job then we don't need the `circuit_json` in the request body?

fake-snippets-api/lib/db/schema.ts
seveibar: We wouldnt use bom_rows in the ui i think, if we do they should probably be parsed

seveibar (approved):
Im a bit afk today, feel free to tag me for subsequent review but generally looks good

PR #813, opened 2025-04-09
feat(routing): add 404 page and handle not found states

ArnavK-09:
Add a new 404 page to handle cases where users navigate to non-existent routes or packages. Implement logic in the ViewPackagePage to detect missing packages and redirect to the 404 page. This improves user experience by providing clear feedback when resources are not found.

# Before (no showing error that snippet not even exists)

image

# After

image

# This also shows 404 when not existent route is asked
image

### DIFF
--- a/src/pages/404.tsx
+++ b/src/pages/404.tsx
@@ -3,7 +3,6 @@ import { Helmet } from "react-helmet"
 import { Header2 } from "@/components/Header2"
 import Footer from "@/components/Footer"
 import { Button } from "@/components/ui/button"
-import { CircuitBoard } from "lucide-react"
 import { PrefetchPageLink } from "@/components/PrefetchPageLink"
 
 export function NotFoundPage() {
--- a/src/pages/404.tsx
+++ b/src/pages/404.tsx
@@ -5,11 +5,13 @@ import Footer from "@/components/Footer"
 import { Button } from "@/components/ui/button"
 import { PrefetchPageLink } from "@/components/PrefetchPageLink"
 
-export function NotFoundPage() {
+export function NotFoundPage({
+  heading = "Page Not Found",
+}: { heading?: string }) {
   return (
     <div className="flex min-h-screen flex-col">
       <Helmet>
-        <title>404 - Page Not Found | tscircuit</title>
+        <title>404 - {heading} | tscircuit</title>
         <meta
           name="description"
           content="The page you're looking for doesn't exist."
@@ -26,7 +28,7 @@ export function NotFoundPage() {
             </div>
           </div>
           <h1 className="text-4xl font-extrabold tracking-tight lg:text-5xl mb-4">
-            Package Not Found
+            {heading}
           </h1>
           <p className="text-xl text-muted-foreground mb-8">
             The page you're looking for doesn't exist or has been moved to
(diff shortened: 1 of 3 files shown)


src/pages/404.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <div className="container px-4 md:px-6 py-12 flex flex-col items-center text-center max-w-3xl">
+          <div className="mb-8 flex flex-col items-center justify-center">
+            <div className="mb-2">
+              <span className="text-3xl font-bold text-white bg-blue-500 px-4 py-2 rounded-md shadow-md inline-block">
+                404
+              </span>
+            </div>
+          </div>
+          <h1 className="text-4xl font-extrabold tracking-tight lg:text-5xl mb-4">
+            Circuit Disconnected
```
seveibar, line 29:
```suggestion
            Package Not Found
```

src/pages/view-package.tsx
```diff
@@ -1,24 +1,57 @@
...
   )
+  useEffect(() => {
+    if (!isLoadingPackageInfo && !isLoadingPackageRelease) {
+      if (
+        packageInfoError?.status === 404 ||
+        packageReleaseError?.status === 404 ||
+        (!packageInfo && !isLoadingPackageInfo) ||
+        (!packageRelease && !isLoadingPackageRelease)
+      ) {
+        setIsNotFound(true)
```
seveibar, line 40: Use early return instead of neated if conditions, you can also use a lot of if conditions, this makes it easier to read

seveibar: *nested

seveibar (commented):
pretty good, see comments

seveibar (approved):

src/pages/view-package.tsx
```diff
@@ -30,15 +30,28 @@ export const ViewPackagePage = () => {
...
+    }
+
+    // Check if package info is missing
+    if (!packageInfo) {
+      setIsNotFound(true)
+      return
+    }
+
+    // Check if package release is missing
+    if (!packageRelease) {
```
seveibar, line 52: You did this backwards, you want to return when there’s no effect/behavior (ie early bail)

ArnavK-09: ohhhhh its too difficult, is this what meant? 

[`cfc7e88` (#813)](https://github.com/tscircuit/tscircuit.com/pull/813/commits/cfc7e889faae66001ea676a75b919b6bd75cf62c)

src/pages/view-package.tsx
```diff
@@ -1,24 +1,57 @@
...
   const { data: packageFiles } = usePackageFiles(
     packageRelease?.package_release_id,
   )
+  useEffect(() => {
+    if (isLoadingPackageInfo || isLoadingPackageRelease) return
+
+    const is404Error =
+      packageInfoError?.status === 404 || packageReleaseError?.status === 404
+    const isMissingData = !packageInfo || !packageRelease
+    console.log(is404Error, isMissingData)
```
seveibar, line 38:
```suggestion
```

seveibar (approved):

src/pages/view-package.tsx
seveibar: yea that's pretty good

PR #820, opened 2025-04-10
feat(routing): add 404 page and handle not found states

ArnavK-09:
Add a new 404 page to handle cases where users navigate to non-existent routes or packages. Implement logic in the ViewPackagePage to detect missing packages and redirect to the 404 page. This improves user experience by providing clear feedback when resources are not found.

image


seveibar (approved):
nice this is a lot cleaner/more obvious logic!

Issue #817, opened 2025-04-09
Allow editing license from ViewPackagePage for package owner


seveibar:
![Image](https://github.com/user-attachments/assets/02f1b181-75ff-49c5-87f3-ba6f6811c232)

- select license from dropdown

ArnavK-09:
/attempt https://github.com/tscircuit/tscircuit.com/issues/817

Anshgrover23:
@ArnavK-09 just close this issue.

PR #823, opened 2025-04-11
feat(package-details): add license management to package details dialog

ArnavK-09:
Introduce license selection and automatic LICENSE file generation for packages. The dialog now supports selecting a license (e.g., MIT) and creates a corresponding LICENSE file when a license is set. This enhances package metadata management and ensures compliance with licensing requirements.

preview: https://discord.com/channels/1233487248129921135/1265314333764878387/1361401475397325083

/claim #817


fake-snippets-api/lib/public-mapping/public-map-package.ts
```diff
@@ -31,7 +32,7 @@ export const publicMapPackage = (internalPackage: {
     latest_package_release_id:
       internalPackage.latest_package_release_id ?? null,
     latest_version: internalPackage.latest_version ?? null,
-    license: internalPackage.latest_license ?? null,
```
imrishabh18, line 34: This is just an alias on selecting, similar to `latest_version` or `latest_package_release_id`. I don't think this will cause any problem and it's best if we follow how the registry is doing it maybe.

Screenshot 2025-04-11 at 15 12 38

fake-snippets-api/lib/public-mapping/public-map-package.ts
```diff
@@ -9,6 +9,7 @@ export const publicMapPackage = (internalPackage: {
   description: string | null
   latest_version?: string | null
   latest_license?: string | null
+  license?: string | null
```
imrishabh18, line 12: Don't need this

fake-snippets-api/routes/api/packages/update.ts
imrishabh18, line 1: I think the license should be updated in the `package_release` and not the `package`. Cause different release can have different license.

cc @seveibar

src/components/dialogs/edit-package-details-dialog.tsx
```diff
@@ -81,11 +112,54 @@ export const EditPackageDetailsDialog = ({
...
         website: website,
+        license: license,
       })
       if (response.status !== 200) {
         console.error("Failed to update package details:", response.data)
         throw new Error("Failed to update package details")
       }
+      // get package files list
+      const packageFiles = []
+      const packageFilesResponse = await axios.post("/package_files/list", {
```
imrishabh18, line 123: Maybe use hooks where it can simplify the code

imrishabh18 (changes requested):
Would like @seveibar's input on this too!

src/components/dialogs/edit-package-details-dialog.tsx
seveibar: yep there already exists a hook for this IIRC

seveibar (commented):
100% agree with @imrishabh18 's review, I'll also add, when the user is changing the license it would be a good idea to show them the LICENSE text, the LICENSE text is modified in github to include the org name so we should just show them the text so they're aware they might have to do that

seveibar (commented):
Can we refactor to “license” from latest_license to match prod, sryyyy

seveibar (approved):

seveibar (approved):

imrishabh18 (approved):

Issue #812, opened 2025-04-09
Fix "page not loaded" intermittent error


seveibar:
After a deploy, there is sometimes an error due to chunks not loading

/bounty $65

ShiboSoftwareDev:
@seveibar can you think of a way to reproduce?

seveibar:
closing since i haven't seen it since we changed some caching headers. sorry ya'll

PR #834, opened 2025-04-12
fix: intermittent-page-load-errors

dekkku:
/claim #812


seveibar (changes requested):
You need to provide some kind of proof this fixes the issue

Issue #837, opened 2025-04-12
Create a underlined text here “show hidden files”


imrishabh18:
The files should be hidden by default, and it should show the hidden directories and files for ex- `/dist/*` and others which are hidden on clicking this.

![Image](https://github.com/user-attachments/assets/16ae51bb-7a9a-47bc-b639-5769de346fe8)

/bounty $5

PR #839, opened 2025-04-13
Created a underlined text “show hidden files” for showing/hiding files

MustafaMulla29:
/claim #837

https://github.com/user-attachments/assets/ee2ae231-89be-46c9-8a40-1c3d90b4f5cf

### DIFF
--- a/src/components/ViewPackagePage/components/tab-views/files-view.tsx
+++ b/src/components/ViewPackagePage/components/tab-views/files-view.tsx
@@ -41,6 +41,7 @@ export default function FilesView({
   onFileClicked,
 }: FilesViewProps) {
   const [activeDir, setActiveDir] = useState("")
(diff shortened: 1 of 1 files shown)


seveibar (changes requested):
text shouldn't be blue

seveibar (changes requested):
looking at the design, it stands out too much

image


Could try this:

image


Btw i'm less likely to review if you request re-review w/o a screenshot AND the vercel deployment isn't deployed. It means i have to come back after it deploys (which i don't get a notification for)

MustafaMulla29:
@seveibar 

https://github.com/user-attachments/assets/c2c66821-7ff0-49f9-a6ee-5e047609ad64

seveibar (approved):

PR #843, opened 2025-04-13
Name updated to Sign up, cause Get started in landing page is present already

imrishabh18:


seveibar (changes requested):
I like the Get Started button up there, why do you want to remove it?

imrishabh18:
image

We already have a `Get started` button, so I thought this just repeats the thing

Issue #866, opened 2025-04-17
Show error when the coordinates for the components are defined but manual-edits is imported


imrishabh18:
Similar to the error warning up on the code editor header when the `manual-edits.json` is not imported but has the edits details in it. We should also show error warning if the code has the `schX`, `schY`, `pcbX` or `pcbY` values but the `manual-edits.json` has the edit details. The warning should show an error message to remove the co-ordinates when using manual edits.

![Image](https://github.com/user-attachments/assets/2a20a481-269e-4d57-87a3-9c6136a19d21)

/bounty $5

seveibar:
/bounty $25 requires a change to core to emit a warning

seveibar:
requires:

- [ ] Warning object defined in circuit-json
- [ ] Warning object emitted from tscircuit/core when there's a position conflict
- [ ] tscircuit.com displays helpful "in your face" error message
- [ ] PCB Viewer displays red box around component where two positions are defined


Do not claim until all 4 issues are resolved or your PR will be closed

ArnavK-09:
- [x] Warning object defined in circuit-json
- [x] Warning object emitted from tscircuit/core when there's a position conflict
- [ ] tscircuit.com displays helpful "in your face" error message
- [ ] PCB Viewer displays red box around component where two positions are defined

PR #868, opened 2025-04-17
Shown error when the coordinates for the components are defined but manual-edits is imported

MustafaMulla29:
/claim #866

### DIFF
--- a/src/components/CodeEditorHeader.tsx
+++ b/src/components/CodeEditorHeader.tsx
@@ -79,6 +79,12 @@ export const CodeEditorHeader = ({
     }
   }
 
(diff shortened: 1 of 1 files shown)


MustafaMulla29:
https://github.com/user-attachments/assets/433c2a95-fe63-43cb-baa2-9ab08bb1efc6

seveibar (changes requested):
it's a core bug not a tscircuit.com bug

seveibar:
The display of the error could be on tscircuit.com, but you need to check for position conflicts inside of core and throw a warning into the circuit json first

Issue #867, opened 2025-04-17
Fix the overflow when hovering over code components


MustafaMulla29:
![Image](https://github.com/user-attachments/assets/b8b979c0-1236-47bc-9997-a64142750b40)

https://github.com/user-attachments/assets/573de3b7-40a3-4beb-adbe-51b422c2d67b

CodeMan62:
/attempt #867
Shouldn't take much time, I'll have a PR up within an hour.

CodeMan62:
Hey @seveibar  this time i have added two images to show what is the changes hope this helps you understand the issue 🙏

PR #870, opened 2025-04-17
Fix tooltip overflow

CodeMan62:
/claim #867
before this pull request the :- 
image
after the pull request it looks like this :-
image

### DIFF
--- a/src/components/CodeEditor.tsx
+++ b/src/components/CodeEditor.tsx
@@ -327,13 +327,94 @@ export const CodeEditor = ({
                 textDecorationColor: "rgba(0, 0, 255, 0.3)",
                 cursor: "pointer",
               },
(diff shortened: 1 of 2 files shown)


seveibar:
sorry @CodeMan62 that doesn't look good

seveibar:
it's good that the spacing is fixed, but we should make it more aesthetically pleasing

CodeMan62:
> it's good that the spacing is fixed, but we should make it more aesthetically pleasing

Let me do this give me some time just a bit

CodeMan62:
@seveibar  
you want it like this or what it means by `aesthetically pleasing`
image

seveibar:
take a look at shadcn, we want stuff to look good like shadcn

CodeMan62:
image
making it look like this at the end

PR #878, opened 2025-04-18
patch: icon color update

ArnavK-09:


seveibar (approved):
wow no dark mode support.... interesting

PR #893, opened 2025-04-20
Restricted edit package name button to owner only

### LINKED ISSUE #871: Restrict 'Edit package name' to owner only
Image

MustafaMulla29:
FIXES #871
/fixes #871


ArnavK-09 (approved):

imrishabh18 (approved):

seveibar:
This is good but FWIW this logic will need to change when we add org support @imrishabh18 @ArnavK-09 , we can probably return the orgs the user is a member of on the user object `account.org_names` most likely

PR #915, opened 2025-04-22
fix: tscircuit.com/trending incomplete features

### LINKED ISSUE #909: fix tscircuit.com/trending incomplete features
Image

/bounty $10

mohan-bee:
closes #909 

/claim #909

https://github.com/user-attachments/assets/3337a142-49f8-4f5b-abe8-5894280634ff


imrishabh18 (changes requested):
I don't see it conveys any sense to have `Last week` and then selecting `Most recent` side by side as a filter option.

It is better to have only one dropdown, which should be having filter options similar to a sub reddit.

cc @seveibar

mohan-bee:
@imrishabh18  can you brief your approach !

PR #921, opened 2025-04-22
Add internal user to filter out the analytics from posthog

imrishabh18:


src/ContextProviders.tsx
```diff
@@ -1,12 +1,59 @@
...
+      return () => clearInterval(checkInterval)
+    }
+
+    const identifyUser = async () => {
+      try {
+        const githubUsername = session?.github_username
+
+        if (isInternalGithubUser(githubUsername)) {
+          posthog.identify(session?.account_id, {
+            is_internal_user: true,
```
seveibar, line 36:
```suggestion
            is_tscircuit_staff: true,
```

src/ContextProviders.tsx
```diff
@@ -1,12 +1,59 @@
...
+          })
+        }
+      } catch (error) {
+        // Error handling silently fails
+      }
```
seveibar, line 41: can you clear the interval or use setTimeout instead? Otherwise we're going to keep identifying the user

src/ContextProviders.tsx
```diff
@@ -1,12 +1,59 @@
...
 import { HelmetProvider } from "react-helmet-async"
+import { useEffect } from "react"
+import { useGlobalStore } from "./hooks/use-global-store"
+import { posthog } from "./lib/posthog"
 
 const queryClient = new QueryClient()
 
+const isInternalGithubUser = (githubUsername?: string | null) => {
+  if (!githubUsername) return false
+  return import.meta.env.VITE_INTERNAL_GITHUB_USERNAMES.split(",").some((internalGithubUsername: string) =>
```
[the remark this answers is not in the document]
seveibar: feel free to add your username as well

Issue #828, opened 2025-04-11
For types, "tscircuit" package should alias to "@tscircuit/core"


seveibar:
![Image](https://github.com/user-attachments/assets/c6cbbc70-1413-4911-a3fe-bdabcbff4c9a)

PR #923, opened 2025-04-22
refactor(CodeEditor): "tscircuit" alias for core && expand ATA run condition to include .ts files

ArnavK-09:
The ATA (Automatic Type Acquisition) initialization was previously limited to `index.tsx`. This change extends it to include all `.tsx` and `.ts` files, ensuring type acquisition works consistently across all TypeScript files.

related to #828


seveibar (approved):
excellent!!! /tip $10

PR #928, opened 2025-04-23
Fix the overflow and tooltip ui when hovering over code components

MustafaMulla29:
/claim #867

image


seveibar (approved):
not sure about the 14px font size but it's ok

can you create an issue to add syntax highlighting and cc me for bounty? tyty

PR #929, opened 2025-04-23
refactor(fake-snippets-api/package_releases): improve atomicity and logic for updating releases

ArnavK-09:
Refactor the update logic for package releases to ensure atomicity when handling `is_latest` updates. The changes simplify the logic by consolidating the update operations and ensuring that updates are only applied when necessary, reducing unnecessary database calls and improving maintainability.

### DIFF
--- a/bun-tests/fake-snippets-api/routes/package_releases/update.test.ts
+++ b/bun-tests/fake-snippets-api/routes/package_releases/update.test.ts
@@ -160,53 +160,6 @@ test("update package release - set is_latest to false when already latest", asyn
   expect(updatedRelease?.is_locked).toBe(true)
 })
 
-test("update package release - set another release to latest and update fields", async () => {
-  const { axios, db } = await getTestServer()
-
-  const pkgRes = await axios.post("/api/packages/create", {
-    name: "testuser/multi-update-test",
-  })
-  const pkgId = pkgRes.data.package.package_id
-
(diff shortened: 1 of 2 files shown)


bun-tests/fake-snippets-api/routes/package_releases/update.test.ts
```diff
@@ -92,18 +92,20 @@ test("update package release - handle is_latest flag", async () => {
...
-  const firstRelease = db.packageReleases.find(
-    (pr) =>
-      pr.package_release_id ===
-      release1.data.package_release.package_release_id,
-  )
-  expect(firstRelease?.is_latest).toBe(false)
+  await axios
+    .post("/api/package_releases/update", {
+      package_release_id: release2.data.package_release.package_release_id,
+      is_latest: true,
```
seveibar, line 98: No change needed, but want to explain the game plan:

The plan in the future is to make it so that the user has to set `is_locked: true` before a package version can be tagged latest. This way, saving a version that wasn't explicitly released will never break people's packages.

We will also support installing `package@unstable`/`package@prerelease`,`package@unstable-latest`, these tags will install the most recent version that is not yet locked

seveibar (approved):

ArnavK-09:
awwwwwwwwwwwwwwwwwwwwww fk, it didnt fixed failing test idk bruh, its passing locally on every try, maybe github issue F, closing

Issue #930, opened 2025-04-23
Add syntax highlighting to the hover tooltip in code editor


MustafaMulla29:
![Image](https://github.com/user-attachments/assets/25da4a02-f09c-4f6e-aec5-0290159ed3f1)

CC @seveibar

seveibar:
/bounty $20

NOTE: We have a preferred syntax highlighter- we use it on the ViewPackagePage for ts code, it's really fast/easy to use relative to other syntax highlighters. I'm not sure what syntax to use for the types stuff, i would find a reference implementation and provide screenshots to show what is industry norm. It might be as simple as just highlighting certain words etc.

PR #934, opened 2025-04-24
Added syntax highlighting to the hover tooltip in code editor

MustafaMulla29:
/claim #930

image

image

### DIFF
--- a/src/components/CodeEditor.tsx
+++ b/src/components/CodeEditor.tsx
@@ -253,8 +253,6 @@ export const CodeEditor = ({
       )
     }
 
-    console.log("IsLoading", isLoading)
-
     // Add TypeScript-specific extensions and handlers
     const tsExtensions =
       currentFile.endsWith(".tsx") || currentFile.endsWith(".ts")
--- a/src/components/CodeEditor.tsx
+++ b/src/components/CodeEditor.tsx
@@ -33,7 +33,6 @@ import React from "@types/react/jsx-runtime"
 import { Circuit, createUseComponent } from "@tscircuit/core"
 import type { CommonLayoutProps } from "@tscircuit/props"
 `
-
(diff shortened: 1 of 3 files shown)


src/components/CodeEditor.tsx
```diff
@@ -256,32 +269,35 @@ export const CodeEditor = ({
...
+              const info = env.languageService.getQuickInfoAtPosition(path, pos)
+              if (!info) return null
+
+              const start = info.textSpan.start
+              const end = start + info.textSpan.length
+              const content = ts.displayPartsToString(info.displayParts || [])
+
+              const dom = document.createElement("div")
+              console.log("content", content)
+              console.log("highlighterRef.current", highlighterRef.current)
```
imrishabh18, line 286:
```suggestion
```

imrishabh18 (changes requested):
See comments!

seveibar (changes requested):
Excellent work, but lets make the usage of the global highlighter singleton in a hook so every component doesnt create one, and helps simplify the code in this component

useShikoHighlighter or something to get the cached highlighter, can use the react query pattern of returning { highlighter, isLoading }

MustafaMulla29:
@seveibar added loading indicator, although not sure if we need it, coz its being used in the useEffect

seveibar (approved):

seveibar:
@MustafaMulla29 indeed you didn't need to put in a loading indicator, the isLoading is just a nice pattern to signal that the user of the hook shouldn't expect the object to be immediately available (it could be unused in this case), sorry that was unclear!

MustafaMulla29:
> @MustafaMulla29 indeed you didn't need to put in a loading indicator, the isLoading is just a nice pattern to signal that the user of the hook shouldn't expect the object to be immediately available (it could be unused in this case), sorry that was unclear!

ok ill update

Issue #953, opened 2025-04-26
If a user goes to /dashboard, show them an empty state that tells them to login


seveibar:
You can still show recent and trending snippets on the left imo, but any good design is acceptable

/bounty $15

PR #955, opened 2025-04-26
If a user goes to /dashboard, show them an empty state that tells them to login

MustafaMulla29:
/claim #953

image

### DIFF
--- a/src/pages/dashboard.tsx
+++ b/src/pages/dashboard.tsx
@@ -116,52 +116,51 @@ export const DashboardPage = () => {
                   </div>
                 </div>
                 <CreateNewSnippetWithAiHero />
-              </>
-            )}
-
-            <h2 className="text-sm font-bold mb-2 text-gray-700 border-b border-gray-200">
-              Your Recent Snippets
-            </h2>
-            {isLoading && <div>Loading...</div>}
-            {mySnippets && (
-              <ul className="space-y-1">
-                {mySnippets.slice(0, 10).map((snippet) => (
-                  <li
-                    key={snippet.snippet_id}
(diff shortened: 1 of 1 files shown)


MustafaMulla29:
@seveibar should i remove the "Your recent snippets" part as well?(when the user is not logged in)

ArnavK-09 (commented):
image
 padding / margin issue on small screens

MustafaMulla29:
@ArnavK-09 
image

ArnavK-09 (approved):
here can be lil margin , then lgtm

image

MustafaMulla29:
image


image

imrishabh18 (changes requested):
image


Also, remove the reference of `Snippets` from the trending and Recet section to `Packages`

MustafaMulla29:
@imrishabh18 
image

imrishabh18 (approved):

seveibar:
killer reviews @ArnavK-09 !!! Appreciated!

seveibar:
@MustafaMulla29 i think it might be a bit better with the border removed and additional margin on the top and bottom. I just attempted to redeploy your branch to vercel so that I can try to make the css modifications. It might be acceptable as is, but I think  we should try it without the border

seveibar:
image

MustafaMulla29:
> @MustafaMulla29 i think it might be a bit better with the border removed and additional margin on the top and bottom. I just attempted to redeploy your branch to vercel so that I can try to make the css modifications. It might be acceptable as is, but I think we should try it without the border

Ok I'll make changes

PR #956, opened 2025-04-26
feat: add manual edit conflict warning handling 

ArnavK-09:
https://github.com/user-attachments/assets/7a51e694-c403-40ae-abff-8421ef4add20

ref #866
/claim #866


src/lib/handleCustomEditsRemoval.ts
```diff
@@ -0,0 +1,59 @@
...
+  updateFileContent: (filename: FileName, content: string) => void,
+  toast: (options: {
+    title: string
+    description: string
+    variant?: "default" | "destructive"
+  }) => void,
+  entrypointFileName: FileName = "index.tsx",
+  isSchematicWarning: boolean = false,
+) => {
+  try {
```
imrishabh18, line 14: Too much stuff happening inside the try block, if possible don't use it. Or just shorten to specific thing you want to handle inside it.

imrishabh18 (commented):
LGTM! Just check the comment

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -390,9 +392,80 @@ export function CodeAndPreview({ pkg }: Props) {
...
+                                    state.pkgFilesWithContent,
+                                    null,
+                                  )?.path,
+                                  isSchematicWarning,
+                                )
+                              }
+                            >
+                              <AlertTriangleIcon className="h-4 w-4" />
+                              <span>Autofix</span>
+                            </button>
```
seveibar, line 460: can we isolate some of this to a separate function? `CodeAndPreview` historically has grown very large

src/lib/handleCustomEditsRemoval.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  updateFileContent(entrypointFileName, modifiedContent)
+
+  toast({
+    title: isSchematicWarning
+      ? "Schematic Coordinates Removed"
+      : "PCB Coordinates Removed",
+    description: `Successfully removed ${isSchematicWarning ? "schematic" : "PCB"} coordinates from components.`,
+    variant: "default",
+  })
```
seveibar, line 47: So this fixes it by removing the pcbX/pcbY, so the button should say "Remove pcbX/pcbY" instead of "Autofix" because we want to be really clear how we're modifying the user's code. Also, "autofix" doesn't make sense because we gave them two options for resolving but 

Also, are you removing ALL the pcbX/pcbY for every component instead of just the component that has a conflict? That would be frustrating if you only wanted to move one component, then autofix removed all of your component positions!

seveibar (commented):
should this be inside RunFrame? I just realized the CLI won't have this helpful warning unless RunFrame does this check?

imrishabh18:
@seveibar Right! I didn't realise that. It should be inside Runframe

ArnavK-09:
> @seveibar Right! I didn't realise that. It should be inside Runframe

Are you smart cli user? \s

seveibar (approved):

seveibar (approved):

Issue #971, opened 2025-04-27
FORK button in own package doesn't make sense.


mohan-bee:
<img width="927" alt="Image" src="https://github.com/user-attachments/assets/2ca678c3-3579-4512-bbb1-0d9bc4171111" />

CC: @seveibar   @imrishabh18

PR #975, opened 2025-04-27
fix: fork button shouldn't be visible on own repo

dekkku:
/claim #971


ArnavK-09:
Issue is already resolved, sorry

PR #977, opened 2025-04-27
Added copy to clipboard option in readme code

### LINKED ISSUE #970: It will be good if we have a option to copy the code in README.md
Image

@imrishabh18

mohan-bee:
Closes #970

### DIFF
--- a/src/components/ViewPackagePage/components/markdown-viewer.tsx
+++ b/src/components/ViewPackagePage/components/markdown-viewer.tsx
@@ -1,8 +1,10 @@
-import React, { useEffect } from "react"
+import React, { useEffect, useState } from "react"
 import Markdown from "react-markdown"
 import remarkGfm from "remark-gfm"
 
 import { useShikiHighlighter } from "@/hooks/use-shiki-highlighter"
+import { Check, Clipboard, Copy, CopyPlus } from "lucide-react"
+import { useToast } from "@/hooks/use-toast"
 
 export default function MarkdownViewer({
   markdownContent,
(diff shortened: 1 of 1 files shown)


src/components/ViewPackagePage/components/markdown-viewer.tsx
```diff
@@ -1,15 +1,39 @@
...
 export default function MarkdownViewer({
   markdownContent,
 }: {
   markdownContent: string
 }) {
   const { highlighter } = useShikiHighlighter()
+  const [copiedCode, setCopiedCode] = useState<string | null>(null)
+  const { toast } = useToast()
+
+  const handleCopy = async (text: string) => {
```
imrishabh18, line 18:
```suggestion
  const copyCodeToClipboard = async (text: string) => {
```

src/components/ViewPackagePage/components/markdown-viewer.tsx
```diff
@@ -1,15 +1,39 @@
...
+    try {
+      await navigator.clipboard.writeText(text)
+      setCopiedCode(text)
+      toast({
+        title: "Copied to clipboard",
+        description: "Your code has been copied successfully.",
+      })
```
imrishabh18, line 25: Even without the toast it is good enough I think. Cause the checked symbol gives the user info that the text is copied. We don't want another toast to convey the message.

src/components/ViewPackagePage/components/markdown-viewer.tsx
```diff
@@ -1,15 +1,39 @@
...
+      setTimeout(() => {
+        setCopiedCode(null)
+      }, 1500)
```
imrishabh18, line 28: I don't think this is a correct practice to use `setTimeout` in this case for animation! Maybe create a custom hook or something else, cause this doesn't make it clear and took me some time to figure out the reason for this timeout, which is used for animation.

cc @seveibar

seveibar: I do see this pattern for copying sometimes, its ok but a comment would help i think eg

// changes check icon back to clipboard

mohan-bee:
@imrishabh18  @seveibar  Ready for review !

PR #978, opened 2025-04-27
fix(CodeEditor): ensure consistent path handling in fsMap

### LINKED ISSUE #948: typings not being updated on package preview
Image

ArnavK-09:
The commit ensures that paths in the `fsMap` are consistently prefixed with a forward slash if they don't already start with one. This prevents potential issues with path resolution in the virtual file system. Additionally, a debug log statement was added to inspect the keys of the `fsMap` for troubleshooting purposes.

##### /claim #948

close #948

### DIFF
--- a/src/components/package-port/CodeEditor.tsx
+++ b/src/components/package-port/CodeEditor.tsx
@@ -136,7 +136,7 @@ export const CodeEditor = ({
         fsMap.set(filename, content)
       })
     })
-    console.info(fsMap.keys())
+
     const system = createSystem(fsMap)
     const env = createVirtualTypeScriptEnvironment(system, [], ts, {
       jsx: ts.JsxEmit.ReactJSX,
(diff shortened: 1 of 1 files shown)


seveibar (commented):
On the backend, we actually intentionally _don't_ use the forward slash prefix, we should keep this consistent everywhere

At first i thought the forward slash prefix was the best solution, but for project-relative paths it interops better with CLI etc. to just not use the forward slash prefix

src/components/package-port/CodeEditor.tsx
```diff
@@ -185,7 +185,7 @@ export const CodeEditor = ({
       delegate: {
         started: () => {
           const manualEditsTypeDeclaration = `
-				  declare module "*.json" {
+				  declare module "manual-edits.json" {
```
seveibar, line 188: this is a nice fix

seveibar (approved):

PR #980, opened 2025-04-27
refactor: rename informManualEditConflicts to toastManualEditConflicts

ArnavK-09:


seveibar (approved):
I would use toastManualEditConflicts as the verb rather than inform, but not bad

PR #985, opened 2025-04-28
fixed rendering of double boards

ShiboSoftwareDev:
/claim  tscircuit/pcb-viewer#247
/closes  tscircuit/pcb-viewer#247

### DIFF
--- a/src/components/CodeAndPreview.tsx
+++ b/src/components/CodeAndPreview.tsx
@@ -202,13 +202,9 @@ export function CodeAndPreview({ snippet }: Props) {
         circuit.add(<Package />)
       `.trim()
     } else {
-      const hasBoard =
-        /<\s*board\s*\/\s*>|<\s*board\s*[^>]*>[\s\S]*?<\/\s*board\s*>/.test(
-          code,
-        )
       entrypointContent = `
         import ${exportName ? `{ ${exportName} as Package }` : "Package"} from "./index.tsx"
-        circuit.add(${hasBoard ? '<Package name="U1" />' : `<board>\n  <Package name="U1" />\n</board>`})
+        circuit.add(<Package name="U1" />)
       `.trim()
     }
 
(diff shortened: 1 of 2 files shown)


src/components/CodeAndPreview.tsx
```diff
@@ -202,13 +202,13 @@ export function CodeAndPreview({ snippet }: Props) {
...
+          code,
+        )
       entrypointContent = `
         import ${exportName ? `{ ${exportName} as Snippet }` : "Snippet"} from "./index.tsx"
-        circuit.add(
-          <board>
-            <Snippet name="U1" />
-          </board>
-        )
+        circuit.add(${hasBoard ? '<Snippet name="U1" />' : `<board>\n  <Snippet name="U1" />\n</board>`})
```
imrishabh18, line 211:
```suggestion
        circuit.add(${hasBoard ? '<Package name="U1" />' : `<board>\n  <Package name="U1" />\n</board>`})
```

We are moving away from the term `Snippets`.

imrishabh18 (commented):
Also, it seems we can get rid of this logic of handling board in the frontend? I remember we are checking for the board in `eval`

cc @seveibar

ShiboSoftwareDev:
I thought this because we want to display the module inside a board even though it doesn't need a board

ShiboSoftwareDev:
@seveibar turning boards into modules should remove the board, wdyt? I think we can merge this for now and create a separate issue

seveibar:
sorry for the delay, yes the logic to handle rendering packages and automatically creating a board if necessary is in eval, and probably is better there as well. tscircuit.com is relatively old so it hasn't shift to that logic.

ShiboSoftwareDev:
@seveibar how do you shift to that logic?
image
I say we merge this to fix the double board for now and create a separate issue for that

seveibar:
@ShiboSoftwareDev i think your claim # is wrong for bounty claim.

Generally I just want to move toward eval because we won't have issues with CLI and tscircuit.com interoperability because the CLI uses eval

seveibar:
eval accepts a fsMap with `circuitWebWorker.executeWithFsMap` IIRC

ShiboSoftwareDev:
> @ShiboSoftwareDev i think your claim # is wrong for bounty claim.

done

seveibar (approved):
accepting for now but we should replace with the prop eval solution soon CC @ArnavK-09

Issue #984, opened 2025-04-28
Users who aren’t logged in receive no response when they click ‘Star’ or ‘Fork.


mohan-bee:
https://github.com/user-attachments/assets/a29bfa81-5779-4a5d-b997-3503a5a227d8

CC: @seveibar  @imrishabh18

imrishabh18:
Good find! Follow the similar approach of Github. Show the hover message and on clicking navigate to the login page

![Image](https://github.com/user-attachments/assets/a531571b-4d8e-497b-8aa0-9c459fefd4a0)

mohan-bee:
@imrishabh18  Let me work on it ! after some hours

mohan-bee:
@imrishabh18  can you assign it to me !

imrishabh18:
We don't follow the issue assignment practice (barring some exceptions), we have seen it generally slow down the project

PR #987, opened 2025-04-29
added a tooltip in fork and star buttons if the user is not logged in

mohan-bee:
Closes: [Issue#984](https://github.com/tscircuit/tscircuit.com/issues/984)

https://github.com/user-attachments/assets/0dc4fd22-ffea-4826-b75f-009af8e4ba40

### DIFF
--- a/src/components/ViewPackagePage/components/package-header.tsx
+++ b/src/components/ViewPackagePage/components/package-header.tsx
@@ -1,16 +1,24 @@
+import React, { useEffect } from "react"
+import { Link } from "wouter"
+
 import { TypeBadge } from "@/components/TypeBadge"
 import { Button } from "@/components/ui/button"
 import { Skeleton } from "@/components/ui/skeleton"
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from "@/components/ui/tooltip"
+import { LockClosedIcon } from "@radix-ui/react-icons"
+import { GitFork, Package, Star } from "lucide-react"
+
 import { useForkPackageMutation } from "@/hooks/use-fork-package-mutation"
 import {
   usePackageStarMutationByName,
   usePackageStarsByName,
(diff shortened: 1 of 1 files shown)


imrishabh18 (changes requested):
You are conditionally rendering the tooltip on hover, but the call is still being made to the backend. I would say disable the Button for these

mohan-bee:
@imrishabh18  I have added a condition in the handleForkClick() if the user is not logged in it will do nothing ..

imrishabh18:
It doesn't give a correct feedback to the user I think. Make the button disabled or maybe navigate to the login page on clicking

mohan-bee:
@imrishabh18  I did something like this 

https://github.com/user-attachments/assets/feae9446-d1eb-4700-85bf-5d096427a563

src/components/ViewPackagePage/components/package-header.tsx
```diff
@@ -167,6 +210,9 @@ export default function PackageHeader({
                 variant="outline"
                 size="sm"
                 onClick={handleForkClick}
+                className={
+                  !isLoggedIn ? "text-gray-400 hover:text-gray-400" : ""
+                }
```
imrishabh18, line 215: You should pass the `isLoggedIn` value to the `disabled` prop and not the className. The style will be handled for  it in the custom Button

imrishabh18 (changes requested):

src/components/ViewPackagePage/components/package-header.tsx
imrishabh18: You will have to wrap the button inside the `Tooltip` for it!

mohan-bee:
@imrishabh18  Disabled button will not trigger a pointer even so I did like this::

https://github.com/user-attachments/assets/00f5f676-f429-4514-bc91-2b15c3f081ae

imrishabh18 (approved):

seveibar (approved):

Issue #989, opened 2025-04-29
Allow build inspection on tscircuit.com


seveibar:
![Image](https://github.com/user-attachments/assets/b33399af-4ce5-4667-a606-a241ab5124d5)

When you hit build failing, we want to open a dialog that shows the circuit json build error for the latest package version. Eventually we'll have a dedicated page, but for now a dialog is fine.

NOTE: We are going to have TWO types of build errors, each will have a different X/check line item

1. Package Transpilation Build
2. Circuit JSON Build

Each of these builds creates files in the `dist` directory, which is visible when you do "Show Hidden Files". Currently many packages are failing and we don't know why, this will help us get to the bottom of it.

There may be some backend changes required

PR #993, opened 2025-04-30
Add build status for the circuit-json

imrishabh18:
Ref - #989 

image


seveibar (approved):
Dialog is perfect, but you should remove the border on the list items under "Releases" and align the icons with the other icons under releases. Also, you should show the CircuitJson build error with a separate button/dialog and the transpilation error with a separate list item/dialog

PR #999, opened 2025-05-02
remove edit beta

seveibar:


ArnavK-09:
shoould we add legacy opt also?

seveibar:
@ArnavK-09 if we had more users maybe, but i dont think anyone is going to complain

PR #1004, opened 2025-05-02
refactor(CodeAndPreview): fix package files not saving for new package

ArnavK-09:
The logic for handling package visibility changes was duplicated in both the `handleSave` function and the `onSave` callback of `PackageVisibilitySettingsDialog`. This commit extracts this logic into a separate `handlePkgVisibilityChange` function to improve code maintainability and reduce redundancy.

### DIFF
--- a/src/components/package-port/CodeAndPreview.tsx
+++ b/src/components/package-port/CodeAndPreview.tsx
@@ -245,7 +245,40 @@ export function CodeAndPreview({ pkg }: Props) {
 
   useWarnUserOnPageChange({ hasUnsavedChanges })
 
+  const handlePkgVisibilityChange = async (isPrivate: boolean) => {
+    setState((prev) => ({ ...prev, lastSavedAt: Date.now() }))
+    const newPackage = await createPackageMutation.mutateAsync({
+      name: `${loggedInUser?.github_username}/${generateRandomPackageName()}`,
+      is_private: isPrivate,
+    })
+
+    if (newPackage) {
+      createRelease(
+        {
+          package_name_with_version: `${newPackage.name}@latest`,
+        },
+        {
+          onSuccess: () => {
+            updatePackageFilesMutation.mutate({
+              package_name_with_version: `${newPackage.name}@latest`,
+              ...newPackage,
+            })
+          },
+        },
+      )
+    }
+  }
+
   const handleSave = async () => {
+    if (!isLoggedIn) {
+      toast({
+        title: "Warning",
+        description: "You must be logged in to save your package.",
+        variant: "destructive",
+      })
+      return
+    }
+
     if (hasUnrunChanges) {
       toast({
         title: "Warning",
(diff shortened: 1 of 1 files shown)


src/components/package-port/CodeAndPreview.tsx
```diff
@@ -245,7 +245,40 @@ export function CodeAndPreview({ pkg }: Props) {
...
+          },
+        },
+      )
+    }
+  }
+
   const handleSave = async () => {
+    if (!isLoggedIn) {
+      toast({
+        title: "Warning",
```
seveibar, line 275:
```suggestion
        title: “Not Logged In",
```

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -245,7 +245,40 @@ export function CodeAndPreview({ pkg }: Props) {
 
   useWarnUserOnPageChange({ hasUnsavedChanges })
 
+  const handlePkgVisibilityChange = async (isPrivate: boolean) => {
+    setState((prev) => ({ ...prev, lastSavedAt: Date.now() }))
+    const newPackage = await createPackageMutation.mutateAsync({
+      name: `${loggedInUser?.github_username}/${generateRandomPackageName()}`,
```
seveibar, line 251: Are you created a new package every time someone changes visibility?

seveibar: I understand this logic was moved but this is terrible

ArnavK-09:
```ts
    if (!pkg) {
      openNewPackageSaveDialog()
      return
    }
```

IT WOULD be only shown when pkg doesnt exist, ie only for new pkgs, same as legacy editor
updated naming

src/components/package-port/CodeAndPreview.tsx
ArnavK-09:
```ts
    if (!pkg) {
      openNewPackageSaveDialog()
      return
    }
```

IT WOULD be only shown when pkg doesnt exist, ie only for new pkgs, same as legacy editor
updated naming

seveibar: I think when i wrote this comment it was called handlePackageVisibilityChange? Maybe not

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -131,10 +131,8 @@ export function CodeAndPreview({ pkg }: Props) {
   const packageType =
     pkg?.snippet_type ?? templateFromUrl?.type ?? urlParams.snippet_type
 
-  const {
-    Dialog: PackageVisibilitySettingsDialog,
-    openDialog: openPackageVisibilitySettingsDialog,
-  } = usePackageVisibilitySettingsDialog()
+  const { Dialog: NewPackageSaveDialog, openDialog: openNewPackageSaveDialog } =
+    usePackageVisibilitySettingsDialog()
```
seveibar, line 136: Wait why is this hook called usePackageVisibilityChange

ArnavK-09: because it was already present in legacy editor, i took it directly from there, and its kinda named correctly it just collects wheter pkg should be public or private
btw this dialog is only being used on before saving any new pkg asking user to wheter save private or public

ArnavK-09:
image

PR #1007, opened 2025-05-02
simplified fsMap that would be passed to runframe

ShiboSoftwareDev:


seveibar (commented):
Tagging @ArnavK-09 to make sure we don't break everything (@ShiboSoftwareDev have you tested on a couple of packages?)

This is definitely the approach I was thinking, we may need to patch eval to make sure it works but yea it's much nicer to have this logic consolidated across all repos evaluating tscircuit

ShiboSoftwareDev:
@seveibar there is a preview

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -288,37 +288,9 @@ export function CodeAndPreview({ pkg }: Props) {
...
-        : `${importStatement}\ncircuit.add(${hasBoard ? '<Package name="U1" />' : `<board>\n  <Package name="U1" />\n</board>`})`
-
     return {
-      ...state.pkgFilesWithContent.reduce(
-        (acc, file) => {
-          acc[file.path] = file.content
-          return acc
-        },
-        {} as Record<string, string>,
-      ),
```
ArnavK-09, line 318: broken directory imports

ArnavK-09: image

ArnavK-09 (changes requested):

ArnavK-09 (approved):
lgtm

seveibar (changes requested):
For the package port, you should be passing all the package files to RunFrame, not just the manual edits and index file content

ShiboSoftwareDev:
@seveibar isn't this what's happening rn? 
image

ShiboSoftwareDev:
I also update the old deprecated file, maybe you thought that was it

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -316,9 +296,8 @@ export function CodeAndPreview({ pkg }: Props) {
         },
         {} as Record<string, string>,
       ),
-      "index.tsx": entryPointCode ?? "// No Default Code Found",
-      "manual-edits.json": state.manualEditsFileContent ?? "{}",
-      "main.tsx": entrypointContent.trim(),
+      "index.tsx": state.code,
+      "manual-edits.json": state.manualEditsFileContent,
```
seveibar, line 300: I think I see the confusion- you're always modifying `index.tsx` even if that's not the file the user is editing? This doesn't make sense, we should edit whatever file the user edited CC @ArnavK-09

ArnavK-09: > I think I see the confusion- you're always modifying `index.tsx` even if that's not the file the user is editing? This doesn't make sense, we should edit whatever file the user edited CC @ArnavK-09 

    Yup notable bug, can be fixed in future pr or this your wish
This pr accomplish the title and lgtm

Also this thing is also needed for cli, so I think we need to find proper way to pass to runframe
#1017

ShiboSoftwareDev: @seveibar I thought you were pointing out a regression

ShiboSoftwareDev: I'll see what I can do

Issue #991, opened 2025-04-30
After login the user is redirected to the default page, rather it should be redirected to the previous page before login in


imrishabh18:
If the user was making some code changes then the code should also persist

mohan-bee:
@imrishabh18  I think this is the issue right !

https://github.com/user-attachments/assets/a8371f3c-8919-41a4-86c6-d5fbfaf1f321

imrishabh18:
Yes!

PR #1015, opened 2025-05-03
Redirect the user to the previous url and persist code changes

MustafaMulla29:
/claim #991

### DIFF
--- a/src/components/package-port/CodeEditor.tsx
+++ b/src/components/package-port/CodeEditor.tsx
@@ -45,6 +45,7 @@ export const CodeEditor = ({
   showImportAndFormatButtons = true,
   onFileContentChanged,
   pkgFilesLoaded,
+  isLoggedIn,
 }: {
   onCodeChange: (code: string, filename?: string) => void
   onDtsChange?: (dts: string) => void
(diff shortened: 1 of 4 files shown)


MustafaMulla29:
@imrishabh18 Used localStorage, not sure if its the correct way

MustafaMulla29:
https://github.com/user-attachments/assets/121a55b2-addb-4b7f-ac2b-d43d16c31d6a

src/components/package-port/CodeEditor.tsx
```diff
@@ -250,6 +252,11 @@ export const CodeEditor = ({
             if (dtsFile?.text && onDtsChange) {
               onDtsChange(dtsFile.text)
             }
+
+            if (!isLoggedIn) {
+              console.log("Saving updated code to localStorage:", newContent)
```
imrishabh18, line 257:
```suggestion
```

src/components/package-port/CodeEditor.tsx
```diff
@@ -414,6 +421,23 @@ export const CodeEditor = ({
...
+      }
+    } else {
+      const saveCode = () => {
+        localStorage.setItem("unsavedCode", code)
+      }
+
+      const interval = setInterval(saveCode, 2000)
+      return () => clearInterval(interval)
+    }
+  }, [code, isLoggedIn])
```
imrishabh18, line 439: Why are you saving every 2 seconds? We should save when the code changes right?

imrishabh18 (changes requested):

MustafaMulla29:
@imrishabh18 fixed

imrishabh18 (approved):

seveibar (changes requested):
This will cause overwriting of user code after the first time there is a redirect

MustafaMulla29:
> This will cause overwriting of user code after the first time there is a redirect

How? It's checking if the doc has changed then it's storing.

PR #1019, opened 2025-05-04
feat: append  mainComponentPath for custom component runs

ArnavK-09:
depend on https://github.com/tscircuit/runframe/pull/542

https://github.com/user-attachments/assets/1f0e9841-198e-46ac-bc54-1cc0f0866229


seveibar (commented):
This is OK but the user should be able to select the current file that they want to preview explicitly, maybe with a button next to the Import ("Preview"?)

image

(btw we should still replace the Import button with the dialog you created in RunFrame which is much much better)

That said, this is good logic to get us started I think.

Note, in the CLI you're supposed to be able to change the active file you're previewing via the RunFrame File menu or possible i little file menu at the top. This makes more sense if you've seen the old CLI frontend

image

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -96,6 +96,23 @@ export function CodeAndPreview({ pkg }: Props) {
       return manualEditsFileFromHook.data.content_text
     else return null
   }, [manualEditsFileFromHook.data])
+  
+  interface CodeAndPreviewState {
```
imrishabh18, line 100: Poor position to define the interface, we should never define it inside function

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -305,8 +334,8 @@ export function CodeAndPreview({ pkg }: Props) {
         },
         {} as Record<string, string>,
       ),
-      "index.tsx": state.code,
-      "manual-edits.json": manualEditsContent,
+      "manual-edits.json": state.manualEditsFileContent,
```
imrishabh18, line 337: Look at my PR merged to this yesterday, we don't want this. I have fixed the manual-edits error

ArnavK-09: yup fixed locally

src/components/package-port/CodeAndPreview.tsx
ArnavK-09: yup fixed in local pr, that commit was just to fix conflictcs, currently wip

seveibar (approved):

seveibar:
The vercel deployment issue due to dep update might be real, maybe an update to circuit-json is required. Annoying sry

PR #1025, opened 2025-05-05
fix: Add destructive toast variant for download error

techmannih:
- Added a toast with variant: "destructive" on download error.
- Displays a user-friendly title and the error message.


imrishabh18 (approved):

seveibar:
just fwiw I vastly prefer react-hot-toast, would love to deprecate this toast entirely

techmannih:
yeah, sound goods, I love react-hot-toast

PR #1028, opened 2025-05-06
After login redirect to previous page

zalabhavy:
/claim https://github.com/tscircuit/tscircuit.com/issues/991

https://github.com/user-attachments/assets/1f983e30-2336-43fc-b8c3-7f80ad9de20f

### DIFF
--- a/src/components/package-port/CodeAndPreview.tsx
+++ b/src/components/package-port/CodeAndPreview.tsx
@@ -244,21 +244,21 @@ export function CodeAndPreview({ pkg }: Props) {
   useEffect(() => {
     if (!isLoggedIn) {
       // Store the current URL before login if not already stored
-      const currentUrl = window.location.href;
+      const currentUrl = window.location.href
       if (!localStorage.getItem("redirectAfterLogin")) {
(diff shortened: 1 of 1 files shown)


Anshgrover23 (changes requested):
@zalabhavy This will cause overwriting of user code after the first time there is a redirect. as told by seve before

zalabhavy:
@Anshgrover23  I have Used
// Ensure redirection does not overwrite user code
      if (window.location.href !== redirectUrl) {
        window.location.replace(redirectUrl)
      } 
Is this not works ?

Anshgrover23:
@zalabhavy by means of overwriting user code is when you redirect to the same url after logging in it does not save previous data written by user so there sholb be an overwritten of code happens...??

zalabhavy:
Yes, @Anshgrover23, when I edit index.tsx in the editor and log in on the same page, my changes are still there after login.

MustafaMulla29 (changes requested):
There is not only index.tsx, user can be editing mutliple files as told by @seveibar

PR #1030, opened 2025-05-06
Created a skeleton for SnippetCard and used it to remove redundant code

MustafaMulla29:


seveibar (approved):
one tweak then it looks good :+1:

src/components/SnippetCardSkeleton.tsx
```diff
@@ -0,0 +1,17 @@
+export const SnippetCardSkeleton = () => {
```
imrishabh18, line 1:
```suggestion
export const PackageCardSkeleton = () => {
```

imrishabh18 (commented):
I think you should use `Package` instead of `Snippet`

src/components/SnippetCardSkeleton.tsx
seveibar: nice catch

src/components/SearchComponent.tsx
```diff
@@ -140,7 +141,7 @@ const SearchComponent: React.FC<SearchComponentProps> = ({
                     className="flex"
                   >
                     <div className="w-12 h-12 overflow-hidden mr-2 flex-shrink-0 rounded-sm">
-                      <img
+                      <OptimizedImage
```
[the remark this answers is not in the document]
MustafaMulla29: @seveibar ok I'll remove the optimizedimage
Btw optimized image component is already present in the codebase(i did not create it)

PR #1032, opened 2025-05-06
refactor(CodeEditorHeader): logic to check if manual imports not used also add support for old spec file 

### LINKED ISSUE #1058: Seeing the error warning by default even for default pacakges
Image

This shouldn't be showing up for packages which are not using the manual-edits

ArnavK-09:
now no default manual edits, only when updated from runframe, 
fixed editor loading state logic
fix #1058

https://github.com/user-attachments/assets/94aebbc3-a6a7-4987-81d2-1ba64913d988

…,

The `checkIfManualEditsImported` function was updated to accept a `file` parameter to check for manual edits in a specific file instead of hardcoding "index.tsx". This ensures the function works correctly for different entry points and file types.

now it checks both

```ts
import manualEdits from "./manual-edits.json";

import manualEdits from "manual-edits.json";
```


src/lib/utils/checkIfManualEditsImported.ts
```diff
@@ -1,6 +1,10 @@
...
+  file: string = "index.tsx",
+) => {
+  if (!files[file]) return false
   const importRegex =
-    /import\s+(?:\*\s+as\s+)?([a-zA-Z_$][\w$]*)\s+from\s+["']\.\/manual-edits\.json["'];?/
-  const hasManualEditsImported = importRegex.test(files["index.tsx"])
-  return files["manual-edits.json"]?.length > 0 && !hasManualEditsImported
+    /import\s+(?:\*\s+as\s+)?([a-zA-Z_$][\w$]*)\s+from\s+["'](?:\.\/)?manual-edits\.json["'];?/
+  const hasManualEditsImported = importRegex.test(files[file])
+  return !!files["manual-edits.json"] && !hasManualEditsImported
```
ArnavK-09, line 9: Use .trim also

seveibar (approved):

imrishabh18 (approved):

PR #1052, opened 2025-05-08
feat: add sign-in prompt for order button

### LINKED ISSUE #1050: Order button should prompt a user to sign in instead of erroring
Image

Cc @imrishabh18

### LINKED ISSUE #1051: Order button doesnt show on mobile

krushnarout:
closes #1050 
depends on https://github.com/tscircuit/runframe/pull/568

https://github.com/user-attachments/assets/8cf9c7a2-0f25-430f-a0b4-6d0279596535


seveibar (changes requested):
Idk feels like we should have a proper sign in dialog with a button here

krushnarout:
> Idk feels like we should have a proper sign in dialog with a button here

@seveibar Done!

Screenshot 2025-05-08 at 11 07 42 AM

imrishabh18 (changes requested):
I don't think this is the right place for this, we should be handling this in runframe repo where the order dialog is.

seveibar:
I agree w @imrishabh18 , also instead of "this action" be specific, "Sign in required to Order" etc

PR #1080, opened 2025-05-12
added schematic-symbols as now it's a peer dependency in core

ShiboSoftwareDev:


ShiboSoftwareDev:
@imrishabh18 can you please merge?

PR #1086, opened 2025-05-12
fixed dropdown not closing once redirected

MustafaMulla29:


MustafaMulla29:
@imrishabh18 @seveibar do you think it's the correct fix?

Anshgrover23 (changes requested):
@MustafaMulla29 bun test failing, also share a video showing the issue fixed.

seveibar:
@MustafaMulla29 can you try this instead https://github.com/shadcn-ui/ui/issues/1971#issuecomment-1817801437

`asChild`

ArnavK-09:
> @MustafaMulla29 bun test failing, also share a video showing the issue fixed.


You can ignore this test failing since it's flaky one, might be fixed in another run
`````````(fail) update package release - handle is_latest flag [15.00ms]`

**Comment by MustafaMulla29** [comment]
> @MustafaMulla29 can you try this instead [shadcn-ui/ui#1971 (comment)](https://github.com/shadcn-ui/ui/issues/1971#issuecomment-1817801437)
> 
> `asChild`

@seveibar tried all the combinations, its not working. I think adding a timeout is the best option here coz the state is not updating immediately after redirection

**Comment by MustafaMulla29** [comment]
https://github.com/user-attachments/assets/5a57b30e-ac4b-4d21-8e63-1b3fe0ec7a2b

@techmannih @Anshgrover23 @seveibar here is the video

**Comment by MustafaMulla29** [comment]
> @MustafaMulla29 thanks for showing this. I think it's still same. Main issue is on click edit online, which doest redirect the dialog where is showing "loading" with circlular loading. cc @imrishabh18 know what the actual issue. Thanks

No I think the issue was, dialog not closing while redirecting and my pr fixes it

**Comment by Anshgrover23** [comment]
@MustafaMulla29 your video is right, yes, that is the main issuethat you have solved but i dont know that if your approach is right or wrong. CC: @imrishabh18 can tell?

**Comment by MustafaMulla29** [comment]
> @MustafaMulla29 your video is right, yes, that is the main issue that you have solved but i don't know that if your approach is right or wrong. CC: @imrishabh18 can tell?

Yes that's what I was asking. I tried all the combinations with `asChild`, `Link` etc but was having the same issue. So decided to add a timeout.

File: src/components/ViewPackagePage/components/main-content-header.tsx [hunk 1]
```diff
@@ -89,9 +89,11 @@ export default function MainContentHeader({
           </DropdownMenuTrigger>
           <DropdownMenuContent align="end" className="w-72">
             <DropdownMenuItem
-              onClick={() => {
+              onSelect={() => {
                 setDropdownOpen(false)
-                setLocation(`/editor?package_id=${packageInfo?.package_id}`)
+                setTimeout(() => {
```
imrishabh18, line 94: It should work without the Timeout

imrishabh18 (changes requested):

src/components/ViewPackagePage/components/main-content-header.tsx
MustafaMulla29: There is still the same issue if timeout is removed.

imrishabh18:
@MustafaMulla29 I think this is a better fix, compared to adding a async timeout in this case. Let me know if you guys are still finding the issue. Otherwise, I will merge it!

https://github.com/tscircuit/tscircuit.com/pull/1095

PR #1087, opened 2025-05-12
Show sign in, when no user not logged in

### LINKED ISSUE #1050: Order button should prompt a user to sign in instead of erroring

imrishabh18:
/close #1050


src/components/ViewPackagePage/components/package-header.tsx
```diff
@@ -40,7 +41,8 @@ export default function PackageHeader({
     packageInfo?.owner_github_username ===
     useGlobalStore((s) => s.session?.github_username)
   const isLoggedIn = useGlobalStore((s) => s.session != null)
-  const { OrderDialog, isOpen, open, close, stage, setStage } = useOrderDialog()
+  const signIn = useSignIn()
+  const { OrderDialog, isOpen, open, close, stage, setStage } = useOrderDialog(signIn)
```
seveibar: This input parameter API is very unclear, when the parameter doesn't logically follow (in this case `useOrderDialog` you would expect maybe `useOrderDialog(order)` not `useOrderDialog(signIn)`) so you have to use named parameters.

Whenever you pass a function to a hook that acts on user input, use the React `on*` convention

```suggestion
  const { OrderDialog, isOpen, open, close, stage, setStage } = useOrderDialog({
    onSignIn: signIn
  })
```

seveibar: also you need to pass a boolean indicating whether or not the user is signed in I think. `{ onSignIn, isLoggedIn }`- idk how you know a user is signed in (others will also be confused!)

seveibar:
RE: the vercel build error, most likely related to pcb-viewer or schematic-viewer being out of date (if I had to guess). I'm a bit conflicted about RunFrame not being properly bundled but for now best to just update. those deps in this repo

imrishabh18:
> RE: the vercel build error, most likely related to pcb-viewer or schematic-viewer being out of date (if I had to guess). I'm a bit conflicted about RunFrame not being properly bundled but for now best to just update. those deps in this repo

But we don't have `pcb-viewer` and `schematic-viewer` as deps in this repo. Probably it could be runframe

PR #1090, opened 2025-05-13
Added Optimized image, so that we show the fallback img when the image doesn't load 

MustafaMulla29:
Before:
image

After:
image


imrishabh18 (changes requested):
OptimisedImage component iirc optimises the image sizing for the image present in the assets and not for dynamic images! 

Try a different approach to add fallback

src/components/PcbImage.tsx
imrishabh18, line 1: remove this component if it's not getting used

PR #1099, opened 2025-05-13
feat: implement redirect to previous page and persist code state after login

### LINKED ISSUE #991: After login the user is redirected to the default page, rather it should be redirected to the previous page before login in

techmannih:
fixes #991 
/claim #991 

https://github.com/user-attachments/assets/7aff2c7f-22ab-4a3e-9466-53235403444a


MustafaMulla29 (changes requested):
@techmannih can you post a video of the fix

imrishabh18 (changes requested):
I don't think this resolves the issue, this has issues like you are never using the `codeState` which is being set. Also, this will overwrite the codeState for different users login

PR #1102, opened 2025-05-13
remove typescript from the bundler

imrishabh18:


seveibar:
image

imrishabh18:
I tried in localhost, and it didn't break it. But yeah the deployed version is broken

PR #1113, opened 2025-05-16
feat: implement file creation 

### LINKED ISSUE #1110: "Add new file" option for package editor
@ArnavK-09 😡😡😡

ArnavK-09:
Resolve #1110

experimental


src/components/FileSidebar.tsx
```diff
@@ -12,6 +23,14 @@ interface FileSidebarProps {
   onFileSelect: (filename: FileName) => void
   className?: string
   fileSidebarState: ReturnType<typeof useState<boolean>>
+  handleCreateFile: (
+    newFileName: string,
+    newFileContent: string,
+    setErrorMessage: (message: string) => void,
+    setIsModalOpen: (isOpen: boolean) => void,
+    onFileSelect: (fileName: string) => void,
+    setNewFileName: (fileName: string) => void,
```
seveibar, line 32: https://github.com/tscircuit/handbook/blob/main/guides/code.md#1-two-parameter-rule

src/components/FileSidebar.tsx
```diff
@@ -98,6 +130,64 @@ const FileSidebar: React.FC<FileSidebarProps> = ({
       >
         <PanelRightOpen />
       </button>
+      <button
+        onClick={() => setIsModalOpen(true)}
+        className="absolute top-2 right-2 text-gray-400 hover:text-gray-600"
+      >
+        <Plus className="w-5 h-5" />
+      </button>
+      <Dialog open={isModalOpen} onOpenChange={setIsModalOpen}>
```
seveibar, line 139: use the dialog-hook convention, it will simplify the implementation and usage

seveibar (commented):
can do more review after vercel deploy

seveibar:
the overlay on this dialog is kind of crazy, can we reduce the overlay.

image

I think it would also be better to do what vscode does and create a new file in the sidebar and actively select, when the user types, it changes the text of the file name until they click away or press enter.

seveibar:
As a short term fix, a more subtle dialog is OK but i think the vs code way is recognizable to people.

ArnavK-09:
> the overlay on this dialog is kind of crazy, can we reduce the overlay.
> 
> image
> 
> I think it would also be better to do what vscode does and create a new file in the sidebar and actively select, when the user types, it changes the text of the file name until they click away or press enter.

ohhhhhhhhhhh man i shouldve seen this comment before, i spend 20 mins debugging why my dialog not working w use dialog ;/

src/components/package-port/CodeEditor.tsx
```diff
@@ -251,16 +253,18 @@ export const CodeEditor = ({
...
+                currentFile,
+                true,
+              )
+              const dtsFile = outputFiles.find((file) =>
+                file.name.endsWith(".d.ts"),
+              )
+              if (dtsFile?.text && onDtsChange) {
+                onDtsChange(dtsFile.text)
+              }
+            } catch {}
```
seveibar, line 267: why is this here? silencing errors is _really really bad_ without an explanation!!!

seveibar (changes requested):
why are we silencing the dts error?

seveibar (approved):

PR #1119, opened 2025-05-17
Fix the github links to redirect to the correct repository 

MustafaMulla29:


ArnavK-09:
Ig that was intentionally done

MustafaMulla29:
> Ig that was intentionally done

So is it the correct link?

imrishabh18:
tscircuit is the main repo which we have, so I agree with Arnav. Check the stars which we have in that repo

PR #1122, opened 2025-05-17
fix(refactor): question text in accordion on small devices

### LINKED ISSUE #1121: FAQ looks bad in small device
Image

techmannih:
fix #1121 

BEFORE
image

AFTER
image


Anshgrover23 (changes requested):
@techmannih can u add a desktop screenshot as well so that we see it didn't broke that.

seveibar (approved):

PR #1143, opened 2025-05-22
patch: remove extra spacing

### LINKED ISSUE #1137: this is also pretty ugly

ArnavK-09:
fix #1137
(ignore lag at starting)

https://github.com/user-attachments/assets/1cdbd849-bddd-4739-878e-cfd5365c1ecc


ArnavK-09:
> screenshot?


https://github.com/user-attachments/assets/262e89fb-66c7-4b2f-9fc0-27e0a0881de0

ArnavK-09:
> > screenshot?
> 
>  Screen.Recording.2025-05-22.203856.mp4

wtf why so lag in recording fk

seveibar (commented):
i think you need to add transition-width to make it smooth

if that doesn't work, you can do `transition-[width,opacity]` or something (you should be able to transition on multiple things

seveibar:
Couple issues (did you test this?) There is still padding and it still does not animate the icon width.

image

seveibar (approved):

PR #1144, opened 2025-05-22
fix: important file preview not updating

### LINKED ISSUE #1129: File's content is not being loaded when you have only one file, unless you click the filename tab
Image

ArnavK-09:
resolves #1129


seveibar (changes requested):
does this actually fix something? Adding dependencies to `useEffect` is the #1 rerender bugs get introduced

seveibar (approved):
note: `importantFiles = []` will cause the effect to fire on every render, should be changed to `importantFiles = emptyArray` so that new arrays aren't being created each time the component is rendered

PR #1162, opened 2025-05-23
feat(refactor): apply consistent capitalization to user-facing strings

techmannih:


Anshgrover23 (changes requested):
@techmannih 
image
 sorry to give u review on this short thing but the login should be always like this see the other apps login screen for reference

PR #1191, opened 2025-06-02
WIP: fix hash function for the preview images

imrishabh18:
Not sure why we were not adding the content_text in the /list response. Was it because of the big payload it can become?

@seveibar

image


seveibar (changes requested):
Yes its because of file size

ArnavK-09:
is this realted to 
image


?

PR #1195, opened 2025-06-04
add missing generate-images command in dev script

Anshgrover23:
fix: not able to generate static images at the time of `bun run dev` locally

# before:

image

# after: 

Screenshot from 2025-06-04 14-05-58


imrishabh18 (approved):
But the screenshot doesn't match with the changes, the landing page should have the Arduino image. It shouldn't be blank

Anshgrover23:
@imrishabh18 forgot to add after image added in descripton.

seveibar (changes requested):
Bad error message for new users, just make a script file

PR #1268, opened 2025-06-12
fix: ui of important file tabs overlapping

ArnavK-09:
# before

image

# after (Scrollable)

image


imrishabh18 (changes requested):
I think there shouldn't be a scrollable view where you have to preview all the files of the package doesn't make sense to me. It should just be the index.ts or the entrypoint file.

ArnavK-09:
> I think there shouldn't be a scrollable view where you have to preview all the files of the package doesn't make sense to me. It should just be the index.ts or the entrypoint file.

fixed in other pr, this pr is only for ui

seveibar (approved):

PR #1273, opened 2025-06-13
add eval as dev dep

seveibar:


ShiboSoftwareDev:
@seveibar let's merge

PR #1286, opened 2025-06-14
Add support for dynamic prefetch pages of packages using and whitelist of prefetchable pages

imrishabh18:


seveibar (approved):
sick dude! make sure to watch carefully after deploy- my only concern with packages is it will trigger a lot of requests and we don't cancel so we might be loading a lot of data. Our api is going to get hammered but that's OK!!!!

PR #1289, opened 2025-06-14
feat: add user profile meta tags

ArnavK-09:
- [x] todo fix conflict bw custom pages and user


seveibar (changes requested):
ehhh we'll hit a github rate limit and our pages will allow slow load if we do this, we have to use a github api key but really we should just be hitting our own backend (we shouldn't rely on third parties just to load our pages)

unrelated but i would move to a separate file because this file is getting crowded

ArnavK-09:
cant seperate due to vercel serveless errors

api/generated-index.js
```diff
@@ -86,6 +95,45 @@ function getHtmlWithModifiedSeoTags({
...
+        github_username: username,
+      },
+    })
+    .json()
+
+  if (!accountResponse?.account?.github_username) {
+    throw new Error("Account not found")
+  }
+  const description = he.encode(
+    `Discover the circuits created by ${accountResponse?.account?.github_username} on tscircuit`,
```
seveibar, line 117:
```suggestion
    `Circuits created by ${accountResponse?.account?.github_username} on tscircuit`,
```

seveibar (changes requested):
Why make the request if youre not using the data, will just cause every page to slow down.

can you use GET if you make the request so that its cached? POST requests are generally not cached

seveibar (approved):

PR #1294, opened 2025-06-15
make LandingPage layout full-width across all sections

abhiiii482:
Updated layout to use full-width containers (w-full) across major sections

image


seveibar:
I dont think we’ll accept this, idk what the problem is youre solving but seems like a design preference

PR #1297, opened 2025-06-16
Move everything to dev deps (fixes issues where @tscircuit/fake-snippets is overriding @tscircuit versions)

seveibar:


seveibar:
CC @ShiboSoftwareDev @imrishabh18  this is going to probably break some packages that use @tscircuit/fake-snippets but I think it's one of the reasons we keep accidentally bundling old versions of some tscircuit dependencies

image

PR #1302, opened 2025-06-16
patch(fake-snippets): sync with prod

ArnavK-09:


seveibar (approved):
i'm not 100% sure about the logic, this looks correct, but we need to prevent bad versions from getting up to the api, there are a lot of "empty string" versions right now

ArnavK-09:
> i'm not 100% sure about the logic, this looks correct, but we need to prevent bad versions from getting up to the api, there are a lot of "empty string" versions right now

It's sync for prod, probably need a look for backend for that

PR #1303, opened 2025-06-16
feat: meta seo tags for pages

ArnavK-09:
WIP


seveibar (approved):
ah interesting concept, how did you test? Looks good but I'm cautious to break prod because this stuff is hard to test

PR #1338, opened 2025-06-21
feat: add Ctrl+B keyboard shortcut to toggle files sidebar

ArnavK-09:
- like vscode


seveibar (approved):
we should have a utility hook like `useHotkey("cmd+b", () => setSideBar(...))` because this takes up a lot of LoC in an important file. Btw cmd can automatically map to ctrl and vice versa etc.

PR #1374, opened 2025-06-26
check for gh user name instead of account id for settings dialog, to make sure it is shown  locally

ArnavK-09:
…


seveibar (approved):
tbh it's a bit jank, especially if we support other login systems besides github later- probably a better solution is to fix the account id

PR #1403, opened 2025-07-03
Discard unsaved changes functionality to avoid reloading

ArnavK-09:
https://github.com/user-attachments/assets/9a78ccd4-cd90-4ecb-9047-370396f0fbc5

### DIFF
--- a/src/components/package-port/EditorNav.tsx
+++ b/src/components/package-port/EditorNav.tsx
@@ -204,15 +204,6 @@ export default function EditorNav({
     },
     { target: window },
   )
-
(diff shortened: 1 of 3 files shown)


src/components/package-port/EditorNav.tsx
```diff
@@ -201,6 +204,15 @@ export default function EditorNav({
     },
     { target: window },
   )
+
+  useHotkeyCombo(
+    "cmd+shift+z",
```
seveibar, line 209: this is the standard hotkey for "redo", is this intentional?

PR #1421, opened 2025-07-07
Updated the footprinter dep to apply the fixed 0603 footprint

MustafaMulla29:


seveibar (approved):

MustafaMulla29:
> which requires it to be updated in core

Ok I'll update it

PR #1460, opened 2025-07-12
Add fallback image for 3D board in PackageCard component

ArnavK-09:
image


seveibar (commented):
can we use the old fallback? The little circuit icon thing?

seveibar (approved):

PR #1463, opened 2025-07-12
Fix generated index route

andrii-balitskyi:
## Summary
- avoid treating `/api/generated-index` as a package request
- no behavioral changes for normal user package routes

## Testing
- `bun test`

------
https://chatgpt.com/codex/tasks/task_b_686e8d29dba8832cbd5becfa2a4696a3


andrii-balitskyi:
@seveibar I think this should fix those recurring requests we've seen to the registry

PR #1474, opened 2025-07-14
feat(CodeEditor): support click to view local files for improts

### LINKED ISSUE #1473: CMD+Click on local file path should open that file
CC @ArnavK-09 

Image

ArnavK-09:
fix #1473


ArnavK-09:
> yes if there's any way to not make the CodeEditor too much bigger that is good too!!

Ye pring in next

PR #1492, opened 2025-07-19
 ui patches to fix annoying stuff 

ArnavK-09:


seveibar (changes requested):
we actually want to link to the API so that people know it's API-accessible. Yes it's a bit weird, but I think it makes sense

seveibar:
If you're downloading JSON you probably want it via API anyway

seveibar (approved):

PR #1509, opened 2025-07-25
fix(CodeEditorHeader): pass correct component name to fetchEasyEDAComponent

ArnavK-09:
…

The component name was hardcoded as "C1" which caused incorrect component fetching. Now it uses the actual component.name from props.


src/components/package-port/CodeEditorHeader.tsx
```diff
@@ -162,7 +162,8 @@ export const CodeEditorHeader: React.FC<CodeEditorHeaderProps> = ({
       if (!session?.token) {
         throw new Error("You need to be logged in to import jlcpcb component")
       }
-      const jlcpcbComponent = await fetchEasyEDAComponent("C1", {
+      console.log(component)
```
seveibar, line 165:
```suggestion
```

PR #1510, opened 2025-07-26
Fix JLCPCB Import

abimaelmartell:
Import using JLCPCB Part Number.

EasyEDA Import uses Part Number

Screenshot 2025-07-25 at 6 32 04 p m

### DIFF
--- a/src/components/package-port/CodeEditorHeader.tsx
+++ b/src/components/package-port/CodeEditorHeader.tsx
@@ -163,12 +163,8 @@ export const CodeEditorHeader: React.FC<CodeEditorHeaderProps> = ({
         throw new Error("You need to be logged in to import jlcpcb component")
       }
 
-      if (!component.partNumber) {
-        throw new Error("JLCPCB component part number is required")
-      }
-
       const jlcpcbComponent = await fetchEasyEDAComponent(
-        component.partNumber,
+        component.partNumber ?? component.name,
         {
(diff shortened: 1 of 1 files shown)


abimaelmartell:
fixes https://github.com/tscircuit/easyeda-converter/issues/293

seveibar (commented):
just CC'ing @ArnavK-09, he patched a fix to this today as well not sure if it's conflicting

seveibar:
@abimaelmartell do you know if you're fixing anything? I re-tested and it's working after @ArnavK-09 's fix

abimaelmartell:
@seveibar 

https://github.com/tscircuit/easyeda-converter/blob/main/lib/websafe/fetch-easyeda-json.ts#L50

```ts

  const bestMatchComponent =
    searchResult.result.lists.lcsc.find(
      (component: any) =>
        component.dataStr.head.c_para["Supplier Part"] === jlcpcbPartNumber,
    ) ?? searchResult.result.lists.lcsc[0]

```

"Supplier Part" is the actual partNumber, we are currently passing the name, eg "RS-03", so we always get first result. It works for most cases, but i noticed inconsistent results

seveibar (approved):

seveibar:
@abimaelmartell awesome

ArnavK-09:
Thanks

PR #1515, opened 2025-07-28
fix: the default tab view on the package view page should be Description

imrishabh18:


seveibar (approved):
seems a bit weird because if ai_description isn't populated i'm not sure what it will show 🤔

PR #1544, opened 2025-08-06
feat: package build  listing endpoints, schema & integration

ArnavK-09:


fake-snippets-api/routes/api/builds/list.ts
```diff
@@ -0,0 +1,99 @@
+import { withRouteSpec } from "fake-snippets-api/lib/middleware/with-winter-spec"
```
seveibar, line 1: it should be `package_builds/list`! Also for `/list` we would never have the option to `include_logs` because it would exceed vercel's response limits, so you'll need `package_build/get`

seveibar (approved):

PR #1562, opened 2025-08-12
feat: add support for downloading GLTF models from Circuit JSON

ArnavK-09:
- Introduced a new function `downloadGltfFromCircuitJson` to handle the conversion and downloading of GLTF models.
- Updated `DownloadButtonAndMenu` component to include options for downloading both GLB and GLTF formats.
- Added `circuit-json-to-gltf` dependency in `package.json` and `bun.lock` for model conversion.


seveibar (approved):
double triple check that this does not increase the bundle size, if it does chunk accordingly via dynamic imports etc.

PR #1624, opened 2025-09-02
updated to react 19 + updated tscircuit packages

ShiboSoftwareDev:
this hasn't been updated for a while


seveibar:
when you're publishing a big dep update, it's good to check that the homepage/"Open Online Example" example completely works

PR #1627, opened 2025-09-02
fix mobiel dropdown view

ArnavK-09:
image


seveibar (approved):
generally we don't want to use `!important` stuff, cuz it gets messy

PR #1694, opened 2025-09-21
fix(FileSidebar): change border style from solid to top-only

### LINKED ISSUE #1677: Double border/misaligned border on file sidebar
Image

CC @ArnavK-09

ArnavK-09:
fix #1677


seveibar (approved):
this part is still pretty bad

image

PR #1711, opened 2025-09-25
Update the req deps to get the offsets fixed

MustafaMulla29:


Abse2001 (changes requested):
can you check why tests are not passing

Abse2001 (approved):

Issue #1696, opened 2025-09-23
Prevent re-download of typescript dependencies on package save (and cache in localStorage)


seveibar:
See how when I save, a million dependencies get downloaded? It's not supposed to do that

https://github.com/user-attachments/assets/590ac431-3b61-4bdb-b806-2774d4b960e0

/bounty $75

seveibar:
I've approved @Sahelisaha04, @bitsbyritik also had the correct approach but @sahelisaha04 did a nice refactor of the file

PR #1716, opened 2025-09-25
fix: updated CodeEditor behavior for issue #1696

saish9901:
/claim #1696

Issue Description
Problem: When clicking "Save" in the code editor, the TypeScript Automatic Type Acquisition (ATA) system was triggering 800+ unnecessary network requests to CDNs like jsdelivr.net to fetch type definitions. This caused:

-Slow save operations
-Network spam in browser dev tools
-Increased CDN costs
-Poor user experience
-Solution Implemented
-Root Cause: ATA was being triggered immediately when code changed during save operations, causing it to fetch type 

https://github.com/user-attachments/assets/e2e00e71-e937-4acf-9a19-142221ac937b

definitions for all imports in the code.

Fix Applied:

-Save State Monitoring: Added tracking of isSaving state and introduced ataSuppressed state
-Request Blocking: Modified ATA's fetcher to block all type definition requests during save operations
-Extended Suppression: Continue blocking ATA requests for 2 seconds after save completes to catch queued requests
-Debounced Recovery: ATA functionality resumes normally after the suppression period

Code Changes:

-Added ataSuppressed state and saveBlockTimeoutRef
-Modified ATA fetcher to return empty responses when isSaving || ataSuppressed
-Updated useEffect to skip ATA calls during suppression period
-Added cleanup for timeout references

Result:

-Network requests reduced from 800+ to just 3 (99.7% reduction)
-Faster save operations
-TypeScript autocomplete still works normally after suppression period
-Clean network tab in dev tools

### DIFF
--- a/src/components/package-port/CodeEditor.tsx
+++ b/src/components/package-port/CodeEditor.tsx
@@ -101,6 +101,8 @@ export const CodeEditor = ({
   const viewRef = useRef<EditorView | null>(null)
   const ataRef = useRef<ReturnType<typeof setupTypeAcquisition> | null>(null)
   const lastReceivedTsFileTimeRef = useRef<number>(0)
+  const saveBlockTimeoutRef = useRef<NodeJS.Timeout | null>(null)
+  const [ataSuppressed, setAtaSuppressed] = useState(false)
   const apiUrl = useApiBaseUrl()
   const [cursorPosition, setCursorPosition] = useState<number | null>(null)
   const [code, setCode] = useState(files[0]?.content || "")
@@ -220,6 +222,15 @@ export const CodeEditor = ({
       typescript: tsModule,
       logger: console,
       fetcher: (async (input: RequestInfo | URL, init?: RequestInit) => {
+        // Block ALL ATA requests during save operations AND extended blocking period
+        if (isSaving || ataSuppressed) {
+          // Return empty response to prevent network requests during save
+          return new Response('{}', { 
+            status: 200, 
+            headers: { 'Content-Type': 'application/json' } 
+          })
+        }
+        
         const registryPrefixes = [
           "https://data.jsdelivr.com/v1/package/resolve/npm/@tsci/",
           "https://data.jsdelivr.com/v1/package/npm/@tsci/",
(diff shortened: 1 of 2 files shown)


saish9901:
> not the right solution, invalidates the type system

removed logic where it was completely blocking all ATA requests-used defer and cache #1696

seveibar (changes requested):
Could be cleaner. Var names are confusing, eg actualFetch is not actually fetch, it's more like uncachedFetchFromCdn

In general, non-inline closures are not acceptable, even if the code you're replacing has closures. The hook logic should be isolated to a file and more clear.

Unless something is saved to localStorage i wouldnt consider it cached, so i wouldnt use that term. Unclear if it really fulfills the bounty of its not actually caching

I dont like the mixed approaches. You have to pick one technique and keep it clean. Eg why are we de bouncing? That changes the UX and no reason is given- also afaik not necessary

saish9901:
> Blocking requests during save isnt the solution (your refactor helped me understand what was happening, but classes are extremely unusual for React state management so i wouldnt use refactor, itbis nice that its clean, but a hook is likely the correct pattern)


-Removed the direct ATA call from the editor setup useEffect
-All ATA calls now go through our useAtaManager hook
-Hook properly blocks calls during isSaving state

Please review the changes @seveibar @imrishabh18

saish9901:
reinitiated format checks @seveibar @imrishabh18

src/components/package-port/CodeEditor.tsx
```diff
@@ -664,9 +640,10 @@ export const CodeEditor = ({
 
     viewRef.current = view
 
-    if (currentFile?.endsWith(".tsx") || currentFile?.endsWith(".ts")) {
-      ata(`${defaultImports}${code}`)
-    }
+    // Don't call ATA directly here - let useAtaManager handle it
+    // if (currentFile?.endsWith(".tsx") || currentFile?.endsWith(".ts")) {
+    //   ata(`${defaultImports}${code}`)
+    // }
```
seveibar, line 646: ghost code...

src/components/package-port/CodeEditor.tsx
```diff
@@ -739,14 +716,8 @@ export const CodeEditor = ({
...
-  useEffect(() => {
-    if (
-      ataRef.current &&
-      (currentFile?.endsWith(".tsx") || currentFile?.endsWith(".ts"))
-    ) {
-      ataRef.current(`${defaultImports}${code}`)
-    }
-  }, [codeImports])
+  // Use hook to manage ATA calls and prevent excessive requests during save
+  useAtaManager(ataRef, code, defaultImports, currentFile, isSaving)
```
seveibar, line 720: breaks the "named parameter" rule. Convert to named parameters

src/lib/ata-fetcher.ts
```diff
@@ -0,0 +1,96 @@
...
+  const lastCodeRef = useRef<string>("")
+
+  useEffect(() => {
+    // Don't trigger ATA during save operations
+    if (isSaving) {
+      // Clear any pending ATA calls
+      if (timeoutRef.current) {
+        clearTimeout(timeoutRef.current)
+        timeoutRef.current = null
+      }
```
seveibar, line 58: there should be no logic based on whether or not the user is saving.....

saish9901:
Done with the requested changes.

Please review the changes. @seveibar  @imrishabh18

src/components/package-port/CodeEditor.tsx
saish9901: Removed ghost code. @seveibar

src/components/package-port/CodeEditor.tsx
saish9901: Used named parameters @seveibar

src/lib/ata-fetcher.ts
saish9901: Save dependency logic removed @seveibar

saish9901:
> - `ataFetch` should be returned by `useAtaManager`
> 
> - don't have logic dependent on save
> 
> - cache or memoize


Made above changes.

Please review changes... @seveibar 
@imrishabh18

PR #1719, opened 2025-09-27
fix: caching of ts dependencies on save

bitsbyritik:
/claim #1696 

Issue: 
TypeScript dependencies were re-downloaded every time a package was saved or the editor loaded, causing unnecessary network requests and slower editor performance. 

This PR: 
- Adds localStorage caching for ATA fetcher responses.
- Restores cached TS/TSX/DTS files on ATA initialization. 
- Caches newly received TypeScript files locally. 
- Uses a TTL to avoid stale cache. 
- Falls back to network fetch if cache is missing or expired.

https://github.com/user-attachments/assets/c122bd50-2459-4771-9115-71c5014022c0

Benefit: Faster editor startup, reduced network usage, and smoother developer experience while keeping dependencies up-to-date automatically.

### DIFF
--- a/src/components/package-port/CodeEditor.tsx
+++ b/src/components/package-port/CodeEditor.tsx
@@ -214,6 +214,9 @@ export const CodeEditor = ({
     const tscircuitAliasDeclaration = `declare module "tscircuit" { export * from "@tscircuit/core"; }`
     env.createFile("tscircuit-alias.d.ts", tscircuitAliasDeclaration)
 
+    const CACHE_PREFIX = "ata-cache:"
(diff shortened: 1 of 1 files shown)


seveibar (commented):
sorry for the delay reviewing this, you can get faster reviews by pinging on the #reviewme channel in discord (https://tscircuit.com/join)

This looks good and i will be testing it. Something that would make it slightly better is moving (as much as possible) the implementation out of the CodeEditor file, since this files seems to grow infinitely all the time

What other people have done is introduced a `useAtaManager` hook which manages the ata instance

thanks, will do a deeper review

PR #1722, opened 2025-09-27
feat(ImportantFilesView): org managers can do tasks

ArnavK-09:


seveibar (commented):
Shouldnt this be for anyone with the can_manage_package permission?

PR #1731, opened 2025-09-27
feat: create pkg under org ui 

ArnavK-09:
image


fake-snippets-api/routes/api/packages/create.ts
```diff
@@ -18,12 +18,37 @@ export default withRouteSpec({
...
   }),
   jsonResponse: z.object({
     package: packageSchema.optional(),
   }),
 })(async (req, ctx) => {
-  const { name, description, is_private, is_unlisted } = req.jsonBody
+  const { name, description, is_private, is_unlisted, org_id } = req.jsonBody
+
+  let targetOrg = null
+  let targetOrgGithubHandle = ctx.auth.github_username
```
seveibar, line 30: This should be inferred from the name of the package, not from auth

seveibar (approved):
it'd also be a good idea to allow the user to change the name in that initial save dialog

also the private public thing takes up so much space it's crazy haha

PR #1735, opened 2025-09-29
fix: prevent confirmation dialog on process of saving new pkg

ArnavK-09:


src/hooks/use-warn-user-on-page-change.ts
```diff
@@ -33,9 +36,9 @@ export default function useWarnUserOnPageChange({
     }
 
     const handlePopState = (event: PopStateEvent) => {
-      if (hasUnsavedChanges) {
+      if (hasUnsavedChanges && Boolean(pkg)) {
         const shouldLeave = window.confirm(
-          "You have unsaved changes. Are you sure you want to leave this page?",
+          "You have unsaved changes. Are you sure you want to leave this page?2",
```
seveibar, line 41:
```suggestion
          "You have unsaved changes. Are you sure you want to leave this page?”
```

PR #1761, opened 2025-10-06
feat: prevent personal orgs from being renamed

### LINKED ISSUE #1758: Personal org shouldn't be allowed to rename on the settings page
It should throw error from the backend and should also be blocked on the client side too.

/bounty $10

Sahelisaha04:
Screenshot From 2025-10-06 14-32-08

/claim #1758

Fixes #1758

### DIFF
--- a/src/pages/organization-settings.tsx
+++ b/src/pages/organization-settings.tsx
@@ -362,7 +362,7 @@ export default function OrganizationSettingsPage() {
                         type="button"
                         variant="outline"
                         onClick={() => form.reset()}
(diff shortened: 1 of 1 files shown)


bun.lock
imrishabh18, line 1: remove the lock file change please, it's not related to the PR

seveibar:
@imrishabh18 we probably should, but we would disable the rename input box (this pr author didnt do this i think) and disable most the fields

the reason to keep it is stuff like deploy keys if we support them in the future. Also upgrading to a paid plan if we support that

PR #1762, opened 2025-10-06
feat: check for orgs avatar

### LINKED ISSUE #1759: Don't fetch the profile images from github for the accounts which doesn't exists in tscirciut
Image

/bounty $5

Ayushjhawar8:
fix #1759


seveibar (changes requested):
If org.github_handle is defined we can show the gh logo, otherwise fall back to letters

PR #1765, opened 2025-10-06
fix: allow public organization pages to be visible without session

### LINKED ISSUE #1764: organization page in not visible without session
Orgs which are public should be visible even when the user is not logged in. 

Image

Image

/bounty $10

DevGajjar28:
fix #1764 
/claim #1764

before:
image

after:
image


imrishabh18 (commented):
I have just added more description to the issue, the organisation `test-1` is still not visible after your fix?

PR #1772, opened 2025-10-07
fix public org visibility when logged out

brymut:
/claim #1764

Fix public org visibility when user is logged out by removing session dependency. 

Screen recording

https://github.com/user-attachments/assets/aad57ee7-dafe-4056-aed3-e3cd75b0f133

### DIFF
--- a/fake-snippets-api/routes/api/accounts/get.ts
+++ b/fake-snippets-api/routes/api/accounts/get.ts
@@ -4,7 +4,7 @@ import { Account, accountSchema } from "fake-snippets-api/lib/db/schema"
 
 export default withRouteSpec({
   methods: ["GET", "POST"],
-  auth: "session",
+  auth: "optional_session",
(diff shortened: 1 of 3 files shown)


brymut:
Can confirm that visibility of public org created is still visible when I log out and go back to visit org page

brymut:
@imrishabh18 is there a way I can see the deployed version? When I recreate locally, solution works.

brymut:
> There is already a PR which did the same changes of what you did in this PR. But take a look at the issue description, I still I can't see the org `test-1` on your deployed version

@imrishabh18 is there a way I can see the deployed version? When I recreate locally, solution works as in the screen recording.

Edit: Also solution slightly different when I check, this only modified the "session" dependency

brymut:
> image

oops, sorry I missed that. Which user does `test-1` belong to?

PR #1820, opened 2025-10-20
refactor: use latest_image_url props for pkg image previews

ArnavK-09:


src/components/PackageCard.tsx
```diff
@@ -82,7 +85,7 @@ export const PackageCard: React.FC<PackageCardProps> = ({
           className={`${imageSize} flex-shrink-0 rounded-md overflow-hidden bg-gray-50 border flex items-center justify-center`}
         >
           <img
-            src={`${baseUrl}/packages/images/${pkg.owner_github_username}/${pkg.unscoped_name}/${availableImages.includes(pkg.default_view || "") ? pkg.default_view : "3d"}.png?fs_sha=${pkg.latest_package_release_fs_sha}`}
+            src={String(previewImageUrl)}
```
seveibar: a bit sketchy, we should be handling the case where it's undefined/null explicitly right?

PR #1846, opened 2025-10-23
feat: new organization creation spec and frontend w/ display name support

ArnavK-09:
- Added normalization for organization names in the create organization API to convert names with spaces and special characters to kebab-case.
- Updated the API schema to enforce name length constraints (5 to 40 characters).
- Modified the organization creation logic to use the normalized name while preserving the original display name.
- Updated tests to verify the normalization functionality.
- Adjusted the organization header component to display the normalized name when available.

### DIFF
--- a/bun-tests/fake-snippets-api/routes/orgs/create.test.ts
+++ b/bun-tests/fake-snippets-api/routes/orgs/create.test.ts
@@ -83,4 +83,4 @@ test("POST /api/orgs/create - should reject invalid org names", async () => {
       expect(error.status).toBe(400)
     }
   }
-})
+}, 10000)


seveibar (changes requested):
this is bad design. The backend should error so that the client obeys the constraints, it should not modify their input. When you create an org you have a display_name and a name

fake-snippets-api/routes/api/orgs/create.ts
```diff
@@ -6,15 +6,25 @@ import { publicOrgSchema } from "fake-snippets-api/lib/db/schema"
...
   methods: ["GET", "POST"],
   commonParams: z.object({
-    name: z.string(),
+    name: z
+      .string()
+      .min(5)
+      .max(40)
+      .regex(
+        /^[a-z0-9-]+$/,
+        "Name must contain only lowercase letters, numbers, and hyphens",
```
seveibar: in github i think underscores are also allowed

seveibar (approved):

PR #1869, opened 2025-10-25
Show toast when session expires

imrishabh18:
## Summary
- add a response interceptor to the shared axios instance
- surface a destructive toast when backend responses return a 401
- include the API base URL in the memoization dependencies so the client refreshes when it changes

## Testing
- not run

------
https://chatgpt.com/codex/tasks/task_b_68fc4ba0e4148327ba9d7d576c87756c

### DIFF
--- a/src/hooks/use-axios.ts
+++ b/src/hooks/use-axios.ts
@@ -17,22 +17,67 @@ export const useAxios = () => {
         : {},
     })
 
-    instance.interceptors.response.use(
-      (response) => response,
-      (error) => {
-        const status = error?.response?.status ?? error?.status
+    // Wrap all instance methods to handle 401 errors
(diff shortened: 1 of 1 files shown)


src/hooks/use-axios.ts
```diff
@@ -15,6 +16,24 @@ export const useAxios = () => {
...
+          })
+        }
+
+        throw error
+      },
+    )
+
     return instance
-  }, [session?.token])
+  }, [session?.token, snippetsBaseApiUrl])
```
seveibar:
```suggestion
  }, [session?.token])
```

seveibar (approved):
see comments, 401s can happen for any unauthorized call, e.g. getting a link to a package you don't own

PR #1880, opened 2025-10-25
refactor: 401 toast

ArnavK-09:
image


seveibar (changes requested):
toasts are always TOP CENTER

PR #1884, opened 2025-10-25
feat: implement organization deletion functionality 

ArnavK-09:


seveibar (approved):
Might be broken on prod, not deleting packages associated with org

PR #1889, opened 2025-10-26
fix: improve unauthorized toast notification handling in useAxios hook

Anshgrover23:
- Unauthorized toast notification appearing multiple times.

- before: 

Screenshot From 2025-10-26 09-40-46

- after: Don't know how to verify, @seveibar Can u check on vercel preview once?

### DIFF
--- a/src/hooks/use-axios.tsx
+++ b/src/hooks/use-axios.tsx
@@ -1,5 +1,5 @@
 import axios from "redaxios"
-import { useMemo } from "react"
+import { useMemo, useRef } from "react"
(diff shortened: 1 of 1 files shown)


Anshgrover23:
@seveibar Can i get a review on this one ?

seveibar:
tbh i'm not sure i care about the spamming, i tagged @ArnavK-09 but personally i just don't think we should care, the user needs to sign out or everything will be broken and the toast is an edge case/backup for when JWT-based detection (not implemented) doesn't work for some reason

PR #1891, opened 2025-10-26
fix: add sessionToken to useAiApi useMemo dependencies

BuddhikaBICT-UoR-FoT-6:
- Prevents stale closures and potential memory leaks
- Ensures Anthropic client is recreated when session changes
- Fixes missing dependency in useMemo hook


imrishabh18 (changes requested):
Was there any issue you faced for which your made these changes?

PR #1892, opened 2025-10-26
fix: remove invalid w-lg class and fix search dropdown width

BuddhikaBICT-UoR-FoT-6:


imrishabh18 (changes requested):
Add screenshots of before and after of what was fixed, what is the second diff file doing in this PR?

PR #1920, opened 2025-10-30
fix wrong image for org

ArnavK-09:


seveibar (approved):
you can save us a future bug and rename username to githubUsername because we're moving to support tscircuit-specific username

PR #1938, opened 2025-11-02
fix: correct source of trending packages in list_trending.ts

### LINKED ISSUE #1934: All star counts for Trending Packages is 1
## Summary

All trending packages on the home page are displayed having 1 star while in reality, it has more. Some having 2, others 5, etc.

Image

Image

rushabhcodes:
/fix #1934 

### Problem
The `list_trending` endpoint was returning packages with 0 stars because it was filtering the wrong array.

### Root Cause
Line 26 was using `ctx.db.packages` instead of `packagesWithStars`, attempting to filter on a `star_count` field that didn't exist yet on the raw package objects.

### Solution
Changed the filter to use `packagesWithStars` which contains the calculated `star_count` field.

### Changes
- Fixed trending packages endpoint to correctly filter packages with stars
- Now properly sorts and returns the top 10 randomized packages from the 50 most-starred packages

### DIFF
--- a/fake-snippets-api/routes/api/packages/list_trending.ts
+++ b/fake-snippets-api/routes/api/packages/list_trending.ts
@@ -24,7 +24,7 @@ export default withRouteSpec({
   })
 
   // Filter out packages with no stars and sort by star count
-  const trendingPackages = ctx.db.packages
+  const trendingPackages = packagesWithStars
     .filter((p) => p.star_count > 0)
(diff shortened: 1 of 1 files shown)


Omar8345:
@rushabhcodes just wanted to help but you are touching unrelated parts of the repo, what you edited is related to fake snippets API, in case of local development to ease testing. that's not related.

PR #1941, opened 2025-11-02
fix: constrain footer link hover area to text only

### LINKED ISSUE #1937: Hover area for footer links is larger than the visible link text
## Summary

When hovering over footer links, the clickable/hover area extends beyond the actual link text. This causes the pointer cursor to appear even when not directly over the text, which may confuse users and make the clickable region feel inconsistent. The issue is likely due to excessive padding or a larger-than-expected anchor element wrapping the text.

Image

RaghavArora14:
Fixes #1937
before and after:

https://github.com/user-attachments/assets/9b20807d-7531-4acd-b2d5-1a81b4f64a12

The issue was that the footer links were placed inside flex column containers (flex flex-col), which caused each link to stretch to the full width of its parent column by default, making the hover effect trigger even when the cursor was far from the actual text or over empty space near the links. This created a poor user experience where the underline would appear unexpectedly. To fix this, I added the w-fit (width: fit-content) Tailwind class to all footer links, which constrains each link's width to only the size of its text content, ensuring the hover effect only triggers when the cursor is directly over the visible text itself.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -51,7 +51,7 @@
         "@tscircuit/3d-viewer": "^0.0.407",
         "@tscircuit/assembly-viewer": "^0.0.5",
         "@tscircuit/create-snippet-url": "^0.0.8",
-        "@tscircuit/eval": "^0.0.447",
+        "@tscircuit/eval": "^0.0.448",
         "@tscircuit/layout": "^0.0.29",
         "@tscircuit/mm": "^0.0.8",
         "@tscircuit/pcb-viewer": "^1.11.218",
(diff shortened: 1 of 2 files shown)


.github/workflows/bundle-size-analysis.yml
```diff
@@ -1,12 +1,6 @@
 name: Bundle Size Analysis
 
-on:
-  pull_request:
-    branches: [main]
-    paths:
-      - 'package.json'
-      - 'bun.lockb'
-      - '.github/workflows/bundle-size-analysis.yml'
```
imrishabh18: Why are you removing this?

imrishabh18 (changes requested):

.github/workflows/bundle-size-analysis.yml
RaghavArora14: seve and shibo asked to remove the on condition, the build check seems to be bugged 
see this https://discord.com/channels/1233487248129921135/1434521457836425306

PR #1955, opened 2025-11-05
Handle session expiration by clearing session when receiving a 401 error

### LINKED ISSUE #1933: User Suddenly Authenticated
## Summary

TL;DR - 2-3 weeks ago I logged in to tscircuit.com with my GitHub account normally, now I open the website, the website looks like if I am authorized, profile picture is visible, but the error toasts prove elsewise, we can take this in matter into 2 issues:

1. Many toast issues for user being unauthenticated
2. ~~Determine sudden cause of issue,~~ while even if auth session expired, it should be handled in a better manner
3. in addition to this issue, public packages will show 404, while in incognito it opens properly, check https://tscircuit.com/seveibar/led-water-accelerometer#3d for example

Image

RaghavArora14:
## Issue Description

Fixes #1933

When a user's session expires or is invalidated on the backend but remains in localStorage, the UI incorrectly shows them as logged in while all API calls fail with 401 errors. This creates:
- Multiple "Unauthorized" toast notifications spamming the screen (one per failed API call)
- UI showing logged-in state while user can't perform authenticated actions
- No automatic recovery - stale session persists until manually cleared

### `src/hooks/use-axios.tsx`
- **Toast deduplication**: Added `has401ToastBeenShown` flag to prevent duplicate toasts when multiple API calls fail simultaneously. Flag resets after 6 seconds.
- **Automatic session cleanup**: On 401 detection, immediately call `setSession(null)` to clear stale session from localStorage, updating UI to logged-out state.
- **Conditional toast**: Only show "Unauthorized" toast when a session exists (prevents errors on public pages).

## Before Video

https://github.com/user-attachments/assets/d2d87095-4a0b-4c2e-9c44-f965b537e9dd

## After Video

https://github.com/user-attachments/assets/e132993b-b26c-4285-8cd3-30d95fa9ac41

Discord Thread:
https://discord.com/channels/1233487248129921135/1435373358442090717

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "@tscircuit/fake-snippets",
@@ -51,7 +52,7 @@
         "@tscircuit/3d-viewer": "^0.0.426",
         "@tscircuit/assembly-viewer": "^0.0.5",
         "@tscircuit/create-snippet-url": "^0.0.8",
-        "@tscircuit/eval": "^0.0.457",
+        "@tscircuit/eval": "^0.0.458",
         "@tscircuit/layout": "^0.0.29",
         "@tscircuit/mm": "^0.0.8",
         "@tscircuit/pcb-viewer": "^1.11.248",
@@ -765,7 +766,7 @@
 
     "@tscircuit/create-snippet-url": ["@tscircuit/create-snippet-url@0.0.8", "", { "dependencies": { "fflate": "^0.8.2" }, "peerDependencies": { "typescript": "^5.0.0" } }, "sha512-VMixgwQRsOXlQGwVh2RZIFLLtsn8YWl2Bht61T26MHNM71A1Wzo5qGZtqcdbVkFnvlA42KmdVVjvxYDvEyWdJw=="],
 
-    "@tscircuit/eval": ["@tscircuit/eval@0.0.457", "", { "peerDependencies": { "@tscircuit/core": "*", "circuit-json": "*", "typescript": "^5.0.0", "zod": "3" } }, "sha512-gG/W5CZmWbLTGnmBjKKjW/vM8oiid4UbEzXuBEw8590L7qdLWHTL5Z5cvKjx8kmTczJupLXEGTLgiqBXMCQN5Q=="],
+    "@tscircuit/eval": ["@tscircuit/eval@0.0.458", "", { "peerDependencies": { "@tscircuit/core": "*", "circuit-json": "*", "typescript": "^5.0.0", "zod": "3" } }, "sha512-EJjTtj5JbHMqyh98pFd8kOwB7eS0EqBs5UqRJP/qy7Iztk5CdVhylWw3t3Nd6NhXtwAW1URTmoOkMVpUWKsoWA=="],
 
     "@tscircuit/featured-snippets": ["@tscircuit/featured-snippets@0.0.1", "", { "peerDependencies": { "typescript": "^5.0.0" } }, "sha512-SNUbCQmyaAaWq7DqqDbYlZkYttbfaObtp5rOheZvlJ2TGYvooECFpB8SzNo06bqKGoIwNjgaAGUTB2DcxdX7ow=="],
 
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@tscircuit/3d-viewer": "^0.0.426",
     "@tscircuit/assembly-viewer": "^0.0.5",
     "@tscircuit/create-snippet-url": "^0.0.8",
-    "@tscircuit/eval": "^0.0.457",
+    "@tscircuit/eval": "^0.0.458",
     "@tscircuit/layout": "^0.0.29",
     "@tscircuit/mm": "^0.0.8",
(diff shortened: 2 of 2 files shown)


seveibar (changes requested):
A 401 doesnt always mean the seasion expired. Check the jwt via a decode (not a verify) to see if its expired

also tell the user you signed them out via a toast

src/hooks/use-axios.tsx
```diff
@@ -31,24 +35,37 @@ export const useAxios = () => {
...
+            {
+              position: "top-center",
+              duration: 5000,
+            },
+          )
+
+          // Reset the flag after a delay so future 401s can show toast again
+          setTimeout(() => {
+            has401ToastBeenShown = false
+          }, 6000)
```
imrishabh18: Not sure if this is the correct pattern, cause all the components will be having the same `setTimeout` which might cause race conditions.

bun.lock
```diff
@@ -3,6 +3,9 @@
   "workspaces": {
     "": {
       "name": "@tscircuit/fake-snippets",
+      "dependencies": {
+        "jwt-decode": "^4.0.0",
```
seveibar: Use jose!

RaghavArora14:
https://github.com/user-attachments/assets/f1325164-4bee-4aa4-ba02-d002bf989c1e

RaghavArora14:
rn this does have two toast messages, one for the sign in prompt and one for the error if youd like i can combine these two together

src/hooks/use-axios.tsx
```diff
@@ -95,5 +80,5 @@ export const useAxios = () => {
     }) as typeof originalPatch
 
     return instance
-  }, [session?.token])
+  }, [session?.token, snippetsBaseApiUrl, setSession, toastLibrary, signIn])
```
seveibar: Dont add unnecessary proeprties to useEffect (we meep useEffect minimal)

src/lib/axios-error-handler.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * Creates an error handler for axios requests that handles 401 errors
+ * by validating the session and showing appropriate toasts
+ */
+export const createAxiosErrorHandler = (
+  session: Store["session"],
+  setSession: (session: Store["session"]) => void,
+  toastLibrary: typeof import("react-hot-toast").default,
```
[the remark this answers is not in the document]
RaghavArora14: done, Fixed by importing [toast] directly in the error handler instead of passing it as a parameter, removing unnecessary coupling

src/hooks/use-axios.tsx
```diff
@@ -95,5 +76,5 @@ export const useAxios = () => {
     }) as typeof originalPatch
 
     return instance
-  }, [session?.token])
+  }, [session?.token, snippetsBaseApiUrl])
```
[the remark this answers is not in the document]
RaghavArora14: oh sorry that's a mistake, you're right should have been removed, fixed it

seveibar (changes requested):
Not what i meant by the breakup- this is messy and an anti-pattern for hook usage.

1. your error handler create function is too broad, it should be specific to 401 handling
2. You cannot pass handlers like setSession or signIn into a function like that-its a bad inverted control flow pattern.

seveibar:
Remember you can create hooks instead of closures-

src/lib/axios-error-handler.tsx
imrishabh18: Move it to some util folder

RaghavArora14:
Changes applied: 
1. Fixed Critical 401 Bug
Session now ONLY clears if JWT token has expired (checks exp claim)
Does NOT clear session on normal 401 authorization errors
Prevents users being signed out when accessing unauthorized resources
2. Fixed Type Issues
Changed NodeJS.Timeout → number (correct for browser environment)
Used window.setTimeout explicitly
3. Fixed Memory Leaks
Added useEffect cleanup to clear timeout refs on unmount
Added signInPromptTimerRef tracking for proper cleanup
4. Fixed Dependency Array
Changed [session, setSession, signIn] → [session]
Removed unused useSignIn import
5. Renamed Function & File for Clarity
isSessionValid → isSessionExpired (expiration check only, not "validity")
is-session-valid.ts → is-session-expired.ts
Function name now matches file name
6. File Organization
Created src/utils/is-session-expired.ts
Deleted src/lib/axios-error-handler.tsx
Updated imports in use-axios-401-handler.tsx

seveibar (approved):

src/hooks/use-axios-401-handler.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const signIn = useSignIn()
+
+  const has401ToastBeenShownRef = useRef(false)
+  const toastResetTimerRef = useRef<NodeJS.Timeout | null>(null)
+
+  const handle401Error = useCallback(() => {
+    if (!session?.token) return
+
+    // Always clear the session on 401 (server rejected the token)
+    setSession(null)
```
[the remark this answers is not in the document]
RaghavArora14: fixed in [ebb81ac](https://github.com/tscircuit/tscircuit.com/pull/1955/commits/ebb81ac23ad2fc43604f16a260a0dc803b947975)

imrishabh18 (approved):

PR #1996, opened 2025-11-13
chore: update @tscircuit/pcb-viewer to v1.11.256

techmannih:


seveibar (approved):
@techmannih we should also make it so that the dependency flow is changed for this repo, then it will stay up to date

image

PR #2106, opened 2025-12-02
fix: introduce zIndexMap in tscircuit.com and increase dialog z-index to prevent orientation cube overlay

### LINKED ISSUE #2104: Orientation cube is blocking the file view
Image

RaghavArora14:
fixes #2104 

- introduces zIndexMap to tscircuit.com
-  sets the z index for the dialogOverlay and dialogContent to 150 and 151

 
image


RaghavArora14:
wip

RaghavArora14:
> instead we should decrease the z-index of the cube, otherwise we will face this issue again and again

image
 
The @tscircuit/runframe CadViewer component renders its container with an inline style of z-index: 1000, which establishes a new stacking context. This means all descendant elements (canvas, orientation cube, etc.) are rendered within this context, regardless of their individual z-index values. Due to CSS stacking context rules, any element outside this container with a z-index lower than 1000 will always render beneath the entire 3D viewer container.
I attempted to resolve this by lowering the container's z-index through CSS overrides and wrapper components, but the inline styles applied by the library take precedence over CSS selectors, making it difficult to reliably override without broad, potentially ambiguous CSS rules that could affect other components.
Therefore, the most reliable and maintainable solution is to increase the dialog's z-index above 1000 (we set it to 1001) to ensure it renders above the 3D viewer's stacking context, rather than attempting to modify the library's internal z-index values.
@rushabhcodes @ShiboSoftwareDev what do you guys think

ShiboSoftwareDev (approved):

PR #2108, opened 2025-12-02
chore: update @tscircuit/3d-viewer to v0.0.448

techmannih:


imrishabh18 (approved):
FYI the runframe needs to have the latest 3d-viewer and then update the runframe here for the latest 3d-viewer to show up on tscircuit.com. Our autoupdate mechanism handles it though

PR #2368, opened 2025-12-29
fix: show all the builds on the View builds page and not the releases

imrishabh18:


src/components/preview/BuildItemRow.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+interface DropdownAction {
+  label: string
+  onClick: (e: React.MouseEvent) => void
+  hidden?: boolean
+}
+
+interface BuildItemRowProps {
+  id: string
```
seveibar: never have naked ids!!!!! add the relevant prefix `package_release_id` etc.

PR #2385, opened 2026-01-02
added format shortcut key in editor

Sahil-Gupta584:
Conneccted to https://github.com/tscircuit/runframe/pull/2253

Preview:

https://github.com/user-attachments/assets/63d57dae-067d-4f96-bfe0-a6b636794e5b


Sahil-Gupta584:
Should i also remove file https://github.com/tscircuit/tscircuit.com/blob/main/src/components/RunButton.tsx ?
Its nowhere getting imported since runbtn is now handled through runframe

seveibar (changes requested):
Too much pollution in an important file

src/lib/utils/handle-format-file.ts
```diff
@@ -0,0 +1,105 @@
...
+  toast,
+}: {
+  files: PackageFile[]
+  currentFile: string | null
+  updateFileContent: (path: string, content: string) => void
+  toast: (toastData: ToasterToast) => string
+}) {
+  if (!window.prettier || !window.prettierPlugins) return
+  if (!currentFile) return
+  try {
```
imrishabh18: try block is too big, make it targeted

PR #2391, opened 2026-01-03
feat: add previewComponentPath as fallback component 

Ayushjhawar8:
Adds support for `previewComponentPath` in `tscircuit.config.json` to allow packages to specify a different component for preview rendering than the main entrypoint.


src/lib/utils/findTargetFile.ts
```diff
@@ -26,6 +26,30 @@ export const findMainEntrypointFileFromTscircuitConfig = (
...
+  files: PackageFile[],
+): PackageFile | null => {
+  const configFile = files.find((file) => file.path === "tscircuit.config.json")
+
+  if (configFile) {
+    try {
+      const config = JSON.parse(configFile.content)
+
+      if (config && typeof config.previewComponentPath === "string") {
+        const previewComponentPath = config.previewComponentPath
```
Ayushjhawar8: Key change here is this read different key from tscircuit.config.json i.e. `previewComponentPath` key

PR #2399, opened 2026-01-04
fix package card  awfull ui

ArnavK-09:


seveibar (approved):
hehe haven't seen this but because of style change i think i know what it does- might have thoughts after preview

seveibar:
haha it's good nice work

Issue #2414, opened 2026-01-05
imrpove builds page ui


ArnavK-09:
ref https://discord.com/channels/1233487248129921135/1265314333764878387/1457493745879486616

PR #2441, opened 2026-01-09
feat(ui): 1: Builds Page

ArnavK-09:
# Before

image

# After
image
image

ref #2414

### DIFF
--- a/src/components/preview/index.tsx
+++ b/src/components/preview/index.tsx
@@ -104,32 +104,9 @@ export const getBuildErrorMessage = (
   if (build.user_code_job_error) {
     if (typeof build.user_code_job_error === "string")
       return build.user_code_job_error
-    if ((build.user_code_job_error as any).message)
-      return (build.user_code_job_error as any).message
+    if ((build.user_code_job_error as { message: string }).message)
+      return (build.user_code_job_error as { message: string }).message
     return "User code job failed"
-  }
-
-  if (build.build_error) {
-    if (typeof build.build_error === "string") return build.build_error
-    if ((build.build_error as any).message)
-      return (build.build_error as any).message
-    return "Build failed"
-  }
-
-  if (build.transpilation_error) {
-    if (typeof build.transpilation_error === "string")
-      return build.transpilation_error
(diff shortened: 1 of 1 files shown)


src/pages/release-builds.tsx
```diff
@@ -56,95 +135,128 @@ export default function ReleaseBuildsPage() {
...
+            <PackageBreadcrumb
+              author={pkg.org_owner_tscircuit_handle || pkg.name.split("/")[0]}
+              packageName={packageName || ""}
+              unscopedName={pkg.unscoped_name}
+              currentPage="builds"
+              releaseVersion={releaseVersion}
+            />
+            <div className="mt-4">
+              <h1 className="text-3xl font-semibold text-gray-900 tracking-tight">
+                Deployments
```
seveibar:
```suggestion
                Builds
```

seveibar (approved):
see comments. GOOD. But _if possible_ try to include a relevant build failure message on each of these lines- i realize this probably requires the backend to parse the logs and look for the relevant error lines, but this is really what the user wants, right now the information is pretty sparse (when the information is this sparse- if we keep it like this without an error message, we should just use a table, there's no need for big blocks when there's no real information being displayed...)

src/components/preview/index.tsx
```diff
@@ -87,3 +95,42 @@ export const StatusIcon = ({ status }: { status: string }) => {
...
+    return "Transpilation failed"
+  }
+
+  if (build.circuit_json_build_error) {
+    if (typeof build.circuit_json_build_error === "string")
+      return build.circuit_json_build_error
+    if ((build.circuit_json_build_error as any).message)
+      return (build.circuit_json_build_error as any).message
+    return "Circuit JSON build failed"
+  }
```
seveibar: I think youre winging it. Most of these are deprecated cc @imrishabh18 , get real data

seveibar: Ie @ArnavK-09 go to prod and find a real failing build

seveibar: Dont use deprecated columns

Issue #2423, opened 2026-01-05
When Importing Component From JLCPCB, It doesn't get added to the imports automatically like importing from @tsci does.


Heinrich-XIAO:
When Importing Component From JLCPCB, It doesn't get added to the imports automatically like importing from @tsci does. Importing from @tsci, it automatically adds `import {} from "@tsci/username.package"`. But when importing from JLCPCB, it just sends me to the file it made.

PR #2478, opened 2026-01-14
feat: #2423 automatically add import for jlcpcb parts as well

Heinrich-XIAO:


seveibar (approved):
see comment, i also added @ArnavK-09 for review

PR #2705, opened 2026-02-10
update circuit-json-to-gerber

ShiboSoftwareDev:


seveibar (approved):
fwiw, for the converters what i'd like to start doing is just doing a dynamic import whenever they're used so that we never need to update them

PR #2786, opened 2026-02-17
Add the badge for the points_to the settings domain page

imrishabh18:
image


seveibar (changes requested):
it should have the release number, e.g. `v0.1.0`, then if the package points to a tag, it can display the tag

PR #2896, opened 2026-03-02
landing 1 base proto review need ,cleanup need

ArnavK-09:


seveibar:
The new landing title for example is really bad, i dont want to argue copy at the same time as design

seveibar:
(noticing improvements, a bit happier ;)

PR #3104, opened 2026-04-08
Fix release/build pages to show updated package domain

mohan-bee:
Motivation: 

When i try to edit the domain in release/build it fails to update in release/build but successfully updates in package description. Also when i try to click on the old domain it says subdomain not found.

https://github.com/user-attachments/assets/38373a0d-5525-41a7-aa1b-a653c868d93a

Release and build pages were reading stale snapshot URLs after domain edits. 

This patch resolves domains from the package’s linked domain list (including latest tag mappings) and falls back only if no live domain exists, so renamed domains show up immediately.


mohan-bee:
@imrishabh18  review pls

PR #3175, opened 2026-04-17
Refactor fabrication and CAD export pipelines to use internal dynamic imports

Abse2001:


src/lib/utils/load-internal-dynamic-modules.ts
```diff
@@ -0,0 +1,38 @@
+import importer from "@tscircuit/internal-dynamic-import"
+import * as circuitJsonToGltf from "circuit-json-to-gltf"
```
seveibar: hmmm this is a problem- you are importing the full module and enlarging the bundle- you have to use `import type` or else you're still bundling that dep

seveibar (approved):
see important comment

seveibar (approved):

PR #3397, opened 2026-05-11
Replace Zod validation with clean handle error

thisisharsh7:
This pr fixes organization creation with uppercase handles showing a raw Zod validation toast by catching the handle validation failure and displaying a inline error.

Tested Locally.

Repro steps -

1. Visit https://tscircuit.com/orgs/new
2. Create an org with uppercase letter 

Before -

https://github.com/user-attachments/assets/98f0176f-b4f8-456d-9de3-3546f61289d3

Screenshot 2026-05-11 at 3 03 07 PM

After - 

https://github.com/user-attachments/assets/dadc6470-2110-447b-920e-1c4ed5ceb713


imrishabh18 (changes requested):
This is not the right place to fix this, it should be a backend fix

thisisharsh7:
changes in the new commit:

- moved org handle validation into the backend route handler
- return a clean error instead of the raw Zod validation message
- kept the frontend unchanged
- updated the existing invalid org handle test to check the structured error response

thisisharsh7:
@imrishabh18 could you please let me know for any changes thanks!

PR #3421, opened 2026-05-13
use resolveAndLoadKicad3dModelFiles for Kicad 3d models in tscircuit.com

mohan-bee:
Updates the KiCad project build path to use resolveAndLoadKicad3dModelFiles from circuit-json-to-kicad@0.0.137. This removes duplicated 3D model path logic from the CLI and keeps model file placement aligned with the KiCad PCB paths generated by the converter.

Motivation:
KiCad project exports currently include the board files but not the 3D model assets, so models can appear missing when the ZIP is opened elsewhere. Packaging the models with the project makes the export self-contained and closer to the CLI behavior.


imrishabh18 (approved):

mohan-bee:
@seveibar @imrishabh18 KiCad export now fetches STEP models before zipping, so can we show a loading state while the download is being prepared?

MustafaMulla29 (approved):

PR #3496, opened 2026-05-25
Use stable keys for org search cards

anil08607:
### Summary
This updates the org search results list to use a stable key for each `OrgCard` instead of the array index.

### What changed
- Replaced `key={i}` with `key={org.org_id || org.tscircuit_handle}` in `src/pages/search.tsx`

### Why
Using the render index as a key can cause React to remount cards when the list order changes. That can lead to unnecessary rerenders and incorrect preservation of component state.

### Impact
- Improves React reconciliation for org search results
- Prevents remounts when search results reorder
- Reduces risk of subtle UI bugs tied to unstable list identity

### Testing
- Verified the org search results now use a stable identity-based key in the render path


imrishabh18 (approved):

MustafaMulla29 (approved):

anil08607:
@MustafaMulla29 merge please

PR #3497, opened 2026-05-25
repro: snippet update database corruption

itisrohit:
adds a reproduction unit test showing that updating a snippet (via `/api/snippets/update`) causes silent database corruption in the mock database client by overwriting unrelated records.

To run the reproduction:
```bash
bun test bun-tests/fake-snippets-api/routes/snippets/update.test.ts
```

### DIFF
--- a/bun-tests/fake-snippets-api/routes/snippets/update.test.ts
+++ b/bun-tests/fake-snippets-api/routes/snippets/update.test.ts
@@ -28,7 +28,9 @@ test("update snippet updates the correct files and does not corrupt the db", asy
     code: "console.log('updated second snippet')",
   })
   expect(updateRes.status).toBe(200)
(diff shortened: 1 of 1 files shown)


itisrohit:
The new test case failed in CI as expected. It demonstrates that updating a snippet (via /api/snippets/update) corrupts unrelated database records. 

This happens because `updateSnippet` uses the index from a filtered subset instead of the global database index when updating `updatedFiles`, causing unrelated files to be overwritten with incorrect/empty data.

imrishabh18 (changes requested):
We are not using the `/snippest/*` api anymore, these are legacy endpoints

itisrohit:
Ah, got it — thanks for the clarification. Since `/snippets/*` is legacy and the newer `/packages/*` / `/package_files/*` flow is not affected by this corruption issue, I’ll go ahead and close this PR.

PR #3522, opened 2026-05-28
Preserve repeated log lines in SSE build logs

anil08607:
## Summary
Remove raw-text deduping from the SSE log stream hook.

## Why
Build logs can legitimately emit the same line more than once. The old dedupe logic dropped those repeated lines.

## Change
- removed client-side dedupe based on `eventType + eventData`

## Result
Repeated log lines now appear correctly in streamed build logs.

### DIFF
--- a/src/hooks/use-sse-log-stream.ts
+++ b/src/hooks/use-sse-log-stream.ts
@@ -10,7 +10,7 @@ export type StreamedLogEntry = {
  * Custom hook to manage Server-Sent Events (SSE) log streaming
  *
  * This hook connects to an SSE endpoint and streams log messages in real-time.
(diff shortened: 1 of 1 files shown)


imrishabh18 (commented):
Why?

imrishabh18:
I remember it was added due to some reason of rendering the sse stream

anil08607:
I removed it because repeated lines are valid build output, and this logic was dropping real logs. If duplicate SSE delivery is still a concern, we should handle that with a stable event id or at the source instead.

imrishabh18:
I see, can you share the circuit where the log lines are being dropped?

PR #3541, opened 2026-05-30
Fix package default view hash

Lathikaa-S:
Fix package default view routing so circuit-based tabs like 3D, PCB, and schematic are not overwritten by `#files` while circuit JSON is still loading.
  
  Motivation: Packages with a saved default view could still open on the Files tab because the page wrote `#files` before circuit JSON finished loading. This makes the default view setting behave as expected when users open a package page.
  
 Before: 
 
 

https://github.com/user-attachments/assets/cb18afe7-4b42-4d4b-9893-c049129d4dcb

After: 

https://github.com/user-attachments/assets/847fcfba-ea4b-4c02-90a1-61862ca69de4


seveibar (approved):
CC @imrishabh18

imrishabh18 (commented):
Can you add a video of before and after?

imrishabh18 (approved):

PR #3552, opened 2026-06-01
Add pinout PNG downloads to the image menu

rushabhcodes:
## What changed
- Added `pinout` support to the PNG download pipeline by rendering through `convertCircuitJsonToPinoutSvg`.
- Added a new `Pinout PNG` item under the Images download submenu.
- Allowed `pinout` to be used as the default image type when the active view is pinout.
- Included the dependency bumps already staged in the branch so the renderer and snippet-url package versions line up with the new support.

## Before
image

## After
image

## Why
Users can already download schematic, PCB, assembly, and 3D images. This change makes pinout diagrams available in the same download flow so pinout-focused packages can be exported consistently.

## Validation
- `bunx biome check src/components/DownloadButtonAndMenu.tsx src/lib/download-fns/download-circuit-png.ts`


seveibar:
Cc @imrishabh18

imrishabh18 (approved):

PR #3600, opened 2026-06-04
Show latest descendant file timestamps for folders in package Files view

rushabhcodes:
## Summary

Add GitHub-style folder timestamps to the package page Files tab.

Folders now display the relative time of the newest visible descendant file, while file rows continue to use the existing `created_at` timestamp from the current API response. This keeps the change frontend-only and avoids any API contract changes.

## What changed

- extract Files tab tree/timestamp derivation into `files-view-utils.ts`
- compute each folder timestamp from the newest visible descendant file
- reuse the same relative date formatter for both files and folders
- preserve existing hidden-file filtering and directory navigation behavior

## Before
image

## After
image

## Current limitation

This is an approximation of `last updated` because the current API only exposes `created_at`, not a true file modification timestamp.

## Follow-up API improvement

If `api.tscircuit.com` later exposes a true file-level `updated_at` timestamp, this UI can become more accurate without changing the overall folder logic.

That API change would improve this by:

- making file rows show the actual last modified time instead of creation time
- making folder rows reflect the newest real edit anywhere inside the folder
- matching GitHub-style `last updated` behavior more closely
- avoiding misleading older timestamps for frequently edited files

The frontend can then switch from `created_at` to `updated_at ?? created_at` with a small follow-up.

### DIFF
--- a/src/components/ViewPackagePage/components/tab-views/files-view-utils.ts
+++ b/src/components/ViewPackagePage/components/tab-views/files-view-utils.ts
@@ -1,0 +1,114 @@
+import type { PackageFile as ApiPackageFile } from "fake-snippets-api/lib/db/schema"
+import { isHiddenFile } from "../../utils/is-hidden-file"
+
+interface PackageFile extends ApiPackageFile {
+  file_content?: string
+  content_text?: string | null
+}
+
+export interface FilesViewDirectoryItem {
+  type: "directory"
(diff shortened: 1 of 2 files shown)


imrishabh18 (commented):
Yes, this is a mostly a backend requirement to add the `updated_at` column for the files.

PR #3764, opened 2026-06-28
Fix editor auth dead-end and unify login redirect handling

rushabhcodes:
## Summary

This PR fixes an auth dead-end in the web editor and unifies login redirect handling across the UI.

Previously, unauthenticated users in the editor saw `Not logged in, can't save`, but that warning was not actionable. On smaller screens, the visible warning could be easier to find than the actual login entry point, leaving users blocked from saving without a clear next step.

This change makes the editor warning directly open the login page and centralizes login redirect path generation so auth entry points consistently preserve the current in-app location.

## Changes

- make the `Not logged in, can't save` warning in the editor clickable
- send users from the editor warning directly to `/login`
- preserve the current path, query string, and hash in the login redirect target
- add shared redirect helpers in `src/lib/utils/handle-redirect.ts`
- reuse the shared login redirect path in:
  - `EditorNav`
  - `HeaderLogin`
  - the 401 auth toast flow in `use-axios`

## Why this matters

- removes a dead-end in the editor save flow
- improves login discoverability, especially on mobile
- keeps users on the same editor context after signing in
- reduces duplicated auth redirect logic across the app

## Root cause

Login redirect path generation was duplicated across multiple auth entry points, and the editor's unauthenticated save warning exposed the blocked state without providing a direct recovery path.

## Validation

- `bun run typecheck`

### DIFF
--- a/src/components/HeaderLogin.tsx
+++ b/src/components/HeaderLogin.tsx
@@ -14,6 +14,7 @@ import { useState } from "react"
 import { useAxios } from "@/hooks/use-axios"
 import { GithubAvatarWithFallback } from "@/components/GithubAvatarWithFallback"
 import { useListUserOrgs } from "@/hooks/use-list-user-orgs"
+import { getLoginPath } from "@/lib/utils/handle-redirect"
 
 export const HeaderLogin = () => {
   const session = useGlobalStore((s) => s.session)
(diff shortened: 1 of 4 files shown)


seveibar (approved):

seveibar:
Dont merge if clicking causes them to lose all their work!

rushabhcodes:
the progress is saved!

PR #3794, opened 2026-07-02
Fix editor package visibility menu to save public/private state from package data

mohan-bee:
### Motivation

The editor page package menu could show or update package visibility independently from the actual package state. This made the editor visibility control feel out of sync with the canonical Package Settings behavior, especially when switching between Public and Private from the editor dropdown.

### Change

- Derive the editor visibility label/checkmark from `pkg.is_private` instead of keeping a separate local visibility source of truth.
- Save editor visibility using the same contract as Package Settings:
  - `public` -> `is_private: false`
  - `private` -> `is_private: true`
- Refetch the package query keys used by the editor after saving visibility.
- Disable the visibility controls and show `Changing...` while the update is in flight.

### Evidence

This matches the Package Settings visibility flow, where visibility is initialized from `packageInfo.is_private` and saved through `is_private = visibility === "private"`.

BEFORE: 

https://github.com/user-attachments/assets/c7469916-2f0f-4d02-a9d8-32bdf60bba15

AFTER: 

https://github.com/user-attachments/assets/a22e98b4-3ebb-4792-9e8c-e93ec2e8ec65


src/components/package-port/EditorNav.tsx
```diff
@@ -635,20 +652,23 @@ export default function EditorNav({
...
-                    : "Public Package"}
+                {isUpdatingVisibility
+                  ? "Changing..."
+                  : canManagePackage
+                    ? isPrivate
+                      ? "Make Public"
+                      : "Make Private"
+                    : isPrivate
+                      ? "Private Package"
+                      : "Public Package"}
```
imrishabh18: No ternary chain! Makes it hard to read

imrishabh18 (approved):
See comment

src/components/package-port/EditorNav.tsx
mohan-bee: um sorry , fixing

PR #3824, opened 2026-07-05
chore: remove unused circuit-json-to-* dependencies

rushabhcodes:
## Summary

Removes 11 `circuit-json-to-*` converter packages from `devDependencies` that are never imported anywhere in the codebase:

`circuit-json-to-3d-png`, `circuit-json-to-bom-csv`, `circuit-json-to-gerber`, `circuit-json-to-gltf`, `circuit-json-to-kicad`, `circuit-json-to-lbrn`, `circuit-json-to-pnp-csv`, `circuit-json-to-readable-netlist`, `circuit-json-to-spice`, `circuit-json-to-step`, `circuit-json-to-tscircuit`

## Why they're unused

- At runtime these converters are loaded from the tscircuit CDN (jscdn with esm.run fallback) via `@tscircuit/internal-dynamic-import`, which always resolves `latest` — the npm pins have no effect on what ships or runs.
- The types don't come from the npm packages either: `src/lib/utils/load-internal-dynamic-modules.ts` declares its own local type stubs for every converter module.
- There are zero imports of any of these packages across `src/`, `fake-snippets-api/`, `scripts/`, and `playwright-tests/`, and no Vite alias or test intercept resolves them locally.

As a side benefit, this stops renovate from opening bump PRs for pins that don't affect the site (the `circuit-*` pattern in `renovate.json` currently auto-updates them).

## Verification

- `bun install` — 11 packages removed, lockfile in sync
- `bun run typecheck` — passes
- `bun run lint` — passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ShiboSoftwareDev:
just to double check see if the exports still work

PR #3874, opened 2026-07-10
Align the fake package-image API with the production backend.

rushabhcodes:
## Summary

- Align the fake package-image API with the production backend.
- Route 3D SVG/PNG and all preview PNG generation through the shared SVG/image service.
- Keep local SVG generation for schematic, PCB, and assembly previews.
- Remove application usage of `circuit-json-to-simple-3d`.
- Update the fake API production URL to `https://registry-api.tscircuit.com`.

## Why

The fake image route was using the obsolete local 3D SVG converter and rasterizing other previews locally, which diverged from the real `api.tscircuit.com` implementation. This keeps local behavior and production behavior consistent while removing the obsolete converter integration.

## Validation

- `bun run typecheck`
- Biome check on the modified files
- `git diff --check`


imrishabh18 (changes requested):
You can't use the fake to call endpoints, use `circuit-to-svg` for generating the svg's

PR #3875, opened 2026-07-10
Fix 3D SVG previews with PNG-backed vectorized rendering

rushabhcodes:
## Summary

Replace the unsupported `circuit-json-to-simple-3d` dynamic loader with the `svg.tscircuit.com` rendering approach:

- Render 3D circuit data to PNG with `circuit-json-to-3d-png`
- Vectorize the PNG into SVG with `@neplex/vectorizer`
- Fall back to an embedded PNG-backed SVG if vectorization fails
- Keep the vectorizer in `devDependencies`

## Root cause

The package-image route attempted to load `circuit-json-to-3d-png` through `@tscircuit/internal-dynamic-import`, but that package is not in the dynamic importer’s supported-module registry, causing 3D SVG requests to return HTTP 500.

The route now imports the renderer directly, matching the server-side approach used by `svg.tscircuit.com`.

## Validation

- Added a focused 3D SVG route regression test
- `bun test bun-tests/fake-snippets-api/routes/packages/images.test.ts bun-tests/fake-snippets-api/routes/packages/3d-images.test.ts`
- `bun run typecheck`
- Biome checks and `git diff --check`


fake-snippets-api/routes/api/packages/images/[owner_github_username]/[unscoped_name]/[view_format].ts
```diff
@@ -95,15 +97,30 @@ export default withRouteSpec({
...
   } else if (outputType === "3d") {
-    const { convertCircuitJsonToSimple3dSvg } =
-      await loadCircuitJsonToSimple3d()
-    svg = await convertCircuitJsonToSimple3dSvg(circuit_json, {
-      background: {
-        color: "#fff",
-        opacity: 0.0,
-      },
-      defaultZoomMultiplier: 1.1,
+    const pngBytes = await renderCircuitJsonTo3dPng(circuit_json, {
```
Comment (seveibar) [inline fake-snippets-api/routes/api/packages/images/[owner_github_username]/[unscoped_name]/[view_format].ts]: honestly we should just remove 3d svgs entirely- all of this is a hack and the original implementation was also quite bad

seveibar (commented):
let's just get rid of 3d svgs

fake-snippets-api/routes/api/packages/images/[owner_github_username]/[unscoped_name]/[view_format].ts
Comment (rushabhcodes) [reply fake-snippets-api/routes/api/packages/images/[owner_github_username]/[unscoped_name]/[view_format].ts]: we will need to remove it from svg.tscircuit.com right

PR #3900, opened 2026-07-13
bump: upgrade circuit-json-to-tscircuit to vv0.0.42

anil08607:
Motivation: I fixed the courtyard rotation issue in the `circuit-json-to-tscircuit` repo about 4 days ago, but it's still not fixed in `tscircuit` or the CLI. I checked it with ChatGPT's help and found that it's a dependency issue.


techmannih (commented):
Please add motivation.

PR #3987, opened 2026-07-21
fix: resolve npm ERESOLVE conflicts and migrate broken biome config

KrishnaX12:
### What this PR does
This PR fixes two critical "plumbing" issues that break the installation and linting pipeline for new developers onboarding to the project.

1. **Fixes `npm install` ERESOLVE crash:** 
   Running `npm install` previously crashed with a hard `ERESOLVE` due to strict peer dependency conflicts (specifically `@tscircuit/3d-viewer` strictly requiring `react@19.1.0` and `jscad-electronics` strictly requiring `@tscircuit/alphabet@^0.0.24`). 
   * **Fix:** Added an `"overrides"` block to `package.json` to forcefully align these strict dependencies with the root project versions, allowing a clean `npm install` without requiring `--legacy-peer-deps`.

2. **Fixes `biome` linter crash:**
   Running `npx @biomejs/biome check src` instantly crashed due to schema errors in `biome.json` (unknown keys like `organizeImports` and `ignore`, and a deprecated `linter` schema).
   * **Fix:** Ran `biome migrate` and bumped `@biomejs/biome` to `^2.5.5` in `package.json` so that the linter can successfully parse the codebase again.

### Testing
- `npm install` now completes successfully with just warnings instead of crashes.
- `npx @biomejs/biome check src` now runs successfully without deserialization crashes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
+    "format-si-unit": "^0.0.7",
     "@anthropic-ai/sdk": "^0.27.3",
     "@babel/preset-react": "^7.25.9",
     "@babel/preset-typescript": "^7.26.0",
@@ -97,7 +98,7 @@
     "@tscircuit/order-dialog": "git+https://github.com/tscircuit/order-dialog.git#55a2839",
     "@tscircuit/pcb-viewer": "1.11.374",
     "@tscircuit/prompt-benchmarks": "^0.0.28",
-    "@tscircuit/props": "^0.0.567",
+    "@tscircuit/props": "^0.0.585",
     "@tscircuit/runframe": "^0.0.2237",
     "@tscircuit/schematic-viewer": "^2.0.74",
     "@types/babel__standalone": "^7.1.7",
@@ -195,5 +196,11 @@
     "zod": "^3.23.8",
     "zustand": "^4.5.5",
     "zustand-hoist": "^2.0.1"
+  },
+  "overrides": {
+    "react": "^19.1.1",
+    "react-dom": "^19.1.1",
+    "@tscircuit/alphabet": "^0.0.25",
+    "@tscircuit/props": "^0.0.585"
   }
 }


package.json
```diff
@@ -195,5 +196,11 @@
     "zod": "^3.23.8",
     "zustand": "^4.5.5",
     "zustand-hoist": "^2.0.1"
+  },
+  "overrides": {
+    "react": "^19.1.1",
+    "react-dom": "^19.1.1",
+    "@tscircuit/alphabet": "^0.0.25",
+    "@tscircuit/props": "^0.0.585"
```
imrishabh18, line 204: why?

imrishabh18 (changes requested):

package.json
KrishnaX12: @imrishabh18 When attempting to install dependencies via standard npm install (without --legacy-peer-deps), npm (v7+) crashes with an ERESOLVE dependency tree resolution error due to strict/experimental peer dependency mismatches across internal packages:

react / react-dom: @tscircuit/3d-viewer strictly requests react: "19.1.0" as a peer dependency, which conflicts with root project's react: "^19.1.1".
@tscircuit/alphabet: jscad-electronics strictly requests @tscircuit/alphabet: "^0.0.24", which conflicts with the root project's ^0.0.25.
@tscircuit/props: Overridden to ^0.0.585 so @tscircuit/core cleanly exports required props (differentialPairProps and internalCircuitProps) without bundling import failures.
Adding "overrides" resolves these peer dependency trees cleanly for developers using npm install!

seveibar: deps like react, props and alphabet should be provided via transient dependencies to tscircuit

also using bun should make sure you don't have this issue- we only support npm for tscircuit consumers but for internal repos we don't bother

KrishnaX12: `@seveibar` that's super valid! I was trying to npm install like an NPC when Bun is lowkey the main character here  Closing this out, thanks for the review!

PR #4046, opened 2026-07-24
Fix user settings to normalize tscircuit handles into valid lowercase slugs

vasu-rs:
Before: 

https://github.com/user-attachments/assets/49227c88-5b51-42a0-92d2-2200f7c0321b

After: 

https://github.com/user-attachments/assets/c518e3ef-b926-4a3b-a9ba-21c714317dcc

### Motivation

User settings submitted tscircuit handles without applying the same normalization behavior used by package settings. Handles containing uppercase letters, spaces, accents, or unsupported characters reached the API and failed with:

> `tscircuit_handle must be lowercase and may only contain lowercase letters, numbers, underscores, and hyphens`

This prevented users from saving otherwise valid handles that could be safely normalized.

For example, `My TSCircuit Handle` is previewed and submitted as `my-tscircuit-handle`.

### User impact

Users can enter natural handle text without encountering an avoidable API validation failure. The settings page now previews and submits a value that conforms to the backend handle requirements.

### Root cause

The user-settings form validated and submitted the raw input, while package settings already normalized names before saving. This caused inconsistent behavior between the two settings pages and allowed invalid uppercase values to reach the account update API.

### DIFF
--- a/src/pages/user-settings.tsx
+++ b/src/pages/user-settings.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react"
+import { useState, useEffect, useMemo } from "react"
 import { useForm } from "react-hook-form"
 import { zodResolver } from "@hookform/resolvers/zod"
 import { z } from "zod"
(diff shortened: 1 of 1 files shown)


mohan-bee (approved):
LGTM

imrishabh18 (changes requested):
Adding space as a tscircuit handle is incorrect, no user name should have a space to be input in the first place

vasu-rs:
> Adding space as a tscircuit handle is incorrect, no user name should have a space to be input in the first place

like this won't accept the username with spaces. it will normalise the username even if the user explictly types with space @imrishabh18

imrishabh18:
I know but that is the wrong feedback to the user, if your username is not having space why let the user enter the space. It was intentional to not allow space

PR #4081, opened 2026-07-26
Fix expired session handling in JLCPCB import

anil08607:
## Motivation

When a user is signed out or their session has expired, the JLCPCB import dialog displays a raw `HTTP 401` error. This technical message does not explain what went wrong or how the user can recover.

The import flow should clearly tell the user that authentication is required and provide an easy way to sign in again.

## Summary

- Validate the user session before opening the JLCPCB import dialog
- Prevent the dialog from displaying raw `HTTP 401` errors
- Show clear messages for missing, invalid, or expired sessions
- Clear invalid or expired sessions from local state
- Provide a clickable sign-in toast for session errors
- Add the session validation route to the fake API
- Add regression coverage for the expired-session flow

## Testing

- `bun run typecheck`
- Auth error-copy unit tests
- Session validation API test
- Playwright expired-session flow test
- Biome checks

### DIFF
--- a/bun-tests/fake-snippets-api/routes/sessions/list.test.ts
+++ b/bun-tests/fake-snippets-api/routes/sessions/list.test.ts
@@ -1,0 +1,11 @@
+import { expect, test } from "bun:test"
+import { getTestServer } from "bun-tests/fake-snippets-api/fixtures/get-test-server"
+
+test("POST /api/sessions/list validates an authenticated session", async () => {
+  const { axios } = await getTestServer()
+
+  const response = await axios.post("/api/sessions/list", {})
+
+  expect(response.status).toBe(200)
+  expect(response.data.sessions).toBeArray()
+})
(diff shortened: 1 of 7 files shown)


seveibar:
Do you have proof this fixes it?

anil08607:
@seveibar yes, I reproduced the same flow with an expired session.

Before: selecting a JLCPCB component showed the raw `HTTP 401` error inside the dialog.

After: the import dialog is blocked and the user gets a “Session Expired” message with an option to sign in again.

mohan-bee:
> @seveibar yes, I reproduced the same flow with an expired session.
> 
> Before: selecting a JLCPCB component showed the raw `HTTP 401` error inside the dialog.
> 
> After: the import dialog is blocked and the user gets a “Session Expired” message with an option to sign in again.

you can add a before and after video

mohan-bee (commented):
i don't think you have to edit the fake snippets !

PR #4088, opened 2026-07-27
Show precise auth toasts for EasyEDA proxy failures

MustafaMulla29:
## Summary

- bump `@tscircuit/runframe` to `0.0.2291` and `tscircuit` to `0.0.2160`; `@tscircuit/eval` is now `0.0.1074`
- detect only the descriptive EasyEDA proxy error propagated by parts-engine: `EasyEDA proxy request failed: <error_code> (HTTP 401)`
- show `Session Expired` for an authoritative `session_expired` response, or for `invalid_token` only when the token used by that render has an expired JWT `exp`
- show `Sign In Required` for a logged-out `no_token` response and a generic `Authentication Failed` message for other proxy authentication failures
- never clear, replace, or otherwise mutate the stored session token
- keep authentication toasts deduplicated with the `auth-401` toast ID
- add a Playwright browser test covering logged-out, unexpired-invalid-token, and expired-token cases, including assertions that stored sessions remain unchanged

## Testing

- `bun run lint`
- `bun run typecheck`
- `bun run build`
- inline parser/expiry behavior check with Bun
- `bunx playwright test playwright-tests/proxy-auth-toast.spec.ts --list`

Full Playwright execution reaches the dev server but is blocked locally before test code runs because the host macOS sandbox denies Chromium's Mach port registration (`Permission denied (1100)`).


imrishabh18 (changes requested):
`source_part_not_found_warning` could also be present in circuit-json if we don't have the part in jlcsearch. Check for the `/proxy` call and if that returns `401` show the toast

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -309,6 +311,21 @@ export function CodeAndPreview({ pkg, projectUrl, isPackageFetched }: Props) {
...
+                const hadSessionToken = Boolean(sessionToken)
+                if (hadSessionToken) setSession(null)
+                toast({
+                  id: "auth-401",
+                  title: hadSessionToken
+                    ? "Session Expired"
+                    : "Sign In Required",
+                  description: hadSessionToken
+                    ? "Your session has expired. Please sign in again."
+                    : "Please sign in to fetch component data, then run again.",
```
seveibar: you're not actually checking for expiration here

src/components/package-port/CodeAndPreview.tsx
```diff
@@ -309,6 +311,21 @@ export function CodeAndPreview({ pkg, projectUrl, isPackageFetched }: Props) {
               setState((prev) => ({ ...prev, lastRunCode: currentFileCode }))
             }
             onRenderFinished={({ circuitJson }) => {
+              if (hasUnauthorizedSourcePartWarning(circuitJson)) {
+                const hadSessionToken = Boolean(sessionToken)
+                if (hadSessionToken) setSession(null)
```
seveibar: REALLY REALLY BAD!@!!!!!!!!

seveibar: do not manipulate session tokens automatically, this will lead to the WORST BUGS EVER! This is the single best way to sabotage the codebase- clearing session tokens like this is the worst possible bug you can introduce

seveibar (changes requested):
see comments NEVER automatically invalidate tokens without tons of checks and user notification!!!!!! in this case you're not even checking to see if there's an issue with the token, extremely bad bug

src/components/package-port/CodeAndPreview.tsx
imrishabh18, line 1: Move this to a separate file, no need of this to be present in CodeAndPreview

seveibar (approved):

imrishabh18 (approved):

PR #4151, opened 2026-07-30
fix: use correct pin labels for RedLed in usb-c-led-flashlight template

GokulPandi-M:
The usb-c-led-flashlight template previously used `LED.neg` and `LED.pos` which are not present on the `@tsci/seveibar.red-led` component. This caused compilation errors for anyone using the template. This PR updates them to `LED.pin2` and `LED.pin1`.

Before:
Screenshot 2026-07-30 at 10 34 47 PM

After:
Screenshot 2026-07-30 at 10 52 43 PM


mohan-bee (commented):
maybe we can use a default LED here . not sure needs suggestion from @seveibar

PR #4177, opened 2026-07-31
Migrate web editor to Monaco WorkspaceCodeEditor with mobile memory toggle

rushabhcodes:
## What changed

Replaces the CodeMirror-based `CodeEditor` in `CodeAndPreview` with `WorkspaceCodeEditor` from `@tscircuit/monaco-code-editor` (v0.0.22), bringing the workspace UI (file sidebar, breadcrumb symbol navigation, quick-open, cross-file search) to the web editor.

## Mobile memory fix

Monaco's TypeScript worker loads the full compiler plus automatically-acquired `@types`, which OOMs mobile Safari (hard per-tab memory cap). The editor package exposes `enableTypeScriptLanguageService`; this PR gates it so the worker and type acquisition never start on mobile.

The gate uses a dedicated hook (`useEnableMonacoLanguageService`) that **starts `false`** and only flips to `true` once a `min-width: 768px` viewport is confirmed. A conventional `!useIsMobile()`-style check reports `false` (desktop) on the first render, so it would briefly enable the language service and spin the worker up before the effect could correct it — defeating the fix on real devices.

## Changes

- Add `@tscircuit/monaco-code-editor` pinned to `v0.0.22` (git `#9d88fd2`)
- `src/lib/monaco-workers.ts`: register Monaco's editor/ts web workers for Vite (`?worker`) — required or language features silently no-op
- Import the package's precompiled stylesheet in `main.tsx`
- Bump `tsconfig` `lib` to ES2021 (package ships raw TS source using `replaceAll`; `target` unchanged)
- Wire file create/rename/delete and content changes to the existing package-file mutations

## Verification

- `bun run typecheck` passes
- Dev server builds cleanly; editor mounts and, on desktop, the TS worker + type acquisition load (full IntelliSense preserved)

### Not verified
Mobile behavior was confirmed at the source level (v0.0.22 disables `modeConfiguration`, `eagerModelSync`, diagnostics, and ATA when the toggle is off) but **not** on a real <768px device — a desktop narrow viewport can confirm TS is disabled but cannot confirm the iOS OOM is resolved. Testing on an actual iOS device is recommended before merge.


seveibar (approved):

rushabhcodes:
@seveibar can we merge this for now, the ctrl + enter not working bug need a fix in runframe

Issue #4204, opened 2026-08-02
npm proxy advertises dist-tags.latest but empty versions{} for x4132/sht40_ad1b_r2 — package findable in search but uninstallable


Hero988:
### Description

The registry npm proxy advertises a `latest` dist-tag for `@tsci/x4132.sht40_ad1b_r2` but serves an empty `versions` map, so the package can be found but never installed.

```
$ bun add @tsci/x4132.SHT40_AD1B_R2
error: Package "@tsci/x4132.SHT40_AD1B_R2" with tag "latest" not found, but package exists
```

### Steps to reproduce

1. Search the registry for "sht4" — `x4132/SHT40_AD1B_R2` comes back (package search API, and it has a normal package page).
2. `bun add @tsci/x4132.SHT40_AD1B_R2` (or the lowercase name) in any project → the error above.
3. `curl https://npm.tscircuit.com/@tsci%2fx4132.sht40_ad1b_r2` shows the inconsistency directly:

```json
{"name":"@tsci/x4132.sht40_ad1b_r2","dist-tags":{"latest":"0.1.0"}, ..., "versions":{}}
```

The registry API itself knows the release: `package_releases/list` returns version 0.1.0 created 2025-08-26, same timestamp as the package. So this isn't a package that was never published — the release exists but the npm proxy has nothing installable for it (possibly a build/transpile that never completed; `has_transpiled_code` is null on the release).

### Why it hurts

This was the only SHT40 sensor package in the registry — exactly the part I needed for a temp/humidity board. As a user the failure is undebuggable: search says the package exists, the site shows it, and bun's "tag not found, but package exists" gives no hint that the published release simply has no artifact. Either hiding release-less packages from search, showing a "no installable release" banner on the package page, or having the proxy return 404 (instead of a dist-tag pointing at nothing) would all make this a much clearer dead end — or better, a rebuild path so the release becomes installable.

NateSpencerWx:
Would a focused package-page fix be useful here: show an explicit “no installable release” state when the selected release has no transpiled artifact, with a fake-API regression case and Playwright coverage for both unavailable and installable releases?

Before starting, could you confirm whether this scope is still wanted and whether you would offer a fixed US$50 bounty for an accepted, tested PR? I am also open to your contributor sponsorship program, but I understand that the four-contribution minimum and monthly scoring do not guarantee a particular payment.

[...]

PR #4205, opened 2026-08-02
fix(view-package): warn when the viewed release has no installable build

Hero988:
## What

Adds a warning banner to the package page when the release being viewed has no installable artifact — i.e. the registry npm proxy advertises a `latest` dist-tag for it but serves an empty `versions` map, so `tsci add` / `bun add` fail with:

```
error: Package "@tsci/x4132.SHT40_AD1B_R2" with tag "latest" not found, but package exists
```

This implements the "no installable release" banner remedy proposed in #4204 — the part of that issue that can be fixed in this repo (the proxy/rebuild side lives in the registry backend).

## Before / after

| Before (production page for `x4132/SHT40_AD1B_R2`, the package from #4204) | After (this PR, same release state) |
| --- | --- |
| before | after |

Before, the page even renders import/usage instructions with no hint that the package cannot be installed.

## How installability is detected

Surveyed live registry releases (`package_releases/get`) against what the npm proxy actually serves:

| package | `has_transpiled` | `latest_package_build_id` | npm proxy `versions` |
| --- | --- | --- | --- |
| `@tsci/x4132.sht40_ad1b_r2` | `false` | `null` | `{}` — uninstallable |
| `@tsci/mustafamulla29.sparkfun-line-sensor` | `false` | `null` | `{}` — uninstallable |
| `@tsci/abse2001.sparkfun-line-sensor-breakout-qre1113-analog` | `true` | `null` | 1 version — installable |
| `@tsci/seveibar.red-led` | `false` | `c37274a7…` | 2 versions — installable |
| `@tsci/seveibar.usb-c-flashlight` | `false` | `bcaaf975…` | 2 versions — installable |

(Three more packages with builds — `anshgrover23.motion-sensor`, `imrishabh18.uv-sensor`, `imrishabh18.air-quality-sensor` — are all installable; `x4132.co2_temp_humidity_sensor` with no build is not.)

So a release is installable iff it has legacy transpiled output (`has_transpiled`) **or** a package build (`latest_package_build_id`). The banner shows only when the release has neither, and is suppressed while `transpilation_in_progress` or `circuit_json_build_in_progress` is true so it never flashes during a normal publish.

## Tests

- `bun-tests/is-package-release-installable.test.ts` — unit test covering the field combinations from the table above (fails without the new util, passes with it; runs in PR CI).
- `playwright-tests/uninstallable-package-banner.spec.ts` — seeds `testuser/uninstallable-board` (release with `has_transpiled: false`, no build), asserts the banner is visible with the right install command and matches an element snapshot, and asserts no banner appears on an installable package (`testuser/my-test-board`).
- `bun test`: 285 pass / 0 fail · `tsc --noEmit`: clean · `biome format`: clean.


seveibar:
not interested in supporting this

PR #4274, opened 2026-08-07
Update Monaco code editor, it now prevent false Monaco diagnostics while TypeScript definitions load

rushabhcodes:
## Summary

- Update `@tscircuit/monaco-code-editor` to upstream commit `ab0ff41` where it fixes a Monaco editor race where diagnostics run before automatic type acquisition has registered downloaded definitions..
  - Preventing temporary red diagnostics
  - Non-blocking type acquisition
  - Delayed loading feedback
  - Accessibility and reduced-motion support
  - Preserving genuine diagnostics and consumer settings
- Refresh the Bun lockfile entry.

## Validation

- `bun install --frozen-lockfile`
- `bun run typecheck`
- `git diff --check`


seveibar:
??? What does this do? Any description?

MustafaMulla29 (approved):

PR #4303, opened 2026-08-08
Add a draggable divider to resize the code editor and preview in the web editor

rushabhcodes:
## What

Replaces the fixed 50/50 split between the code editor and preview on the
package editor page with a **draggable, resizable divider**, so users can give
each pane the width they need instead of being locked to half the screen.

Before: the editor and preview were hard-coded to `md:w-1/2` each, with no way
to widen the editor for long lines or the preview for detailed PCB/schematic
inspection. After: grab the divider and drag; the layout keeps the split for
the session.

https://github.com/user-attachments/assets/007028d7-5d20-423b-a10b-36df4371fac3

## Why it's a UX improvement

- **Resizable panes.** Drag the divider to rebalance editor vs. preview at any
  ratio; a sensible minimum width keeps either pane from collapsing to an
  unusable sliver, and the minimum is re-enforced when the window is resized
  (not just while dragging).
- **Cleaner affordance.** The drag handle is hidden by default and fades in on
  hover/focus, so the divider stays unobtrusive until you reach for it.
- **Keyboard + screen-reader accessible.** The divider is a real `separator`
  with focus styles and keyboard resizing, rather than a mouse-only element.
- **Responsive.** On mobile the editor pane collapses so the preview fills the
  screen; toggling the preview on/off restores the expected layout every time.

## Implementation notes

- Reuses the existing `react-resizable-panels` wrapper in `ui/resizable.tsx`
  (previously unused) instead of hand-rolling pointer math, so pointer capture,
  min-size clamping, and keyboard support come from a maintained library.
- Dragging is driven by the panel library, not React state, so a resize doesn't
  re-render the Monaco editor or the run-frame tree on every pointer move.
- The min-width `ResizeObserver` only updates state when the computed percentage
  actually changes, avoiding per-frame re-renders of a heavy tree during window
  resizing.
- The Monaco `options` object is hoisted to a stable module constant so the
  editor doesn't reconfigure itself on unrelated re-renders.

## Testing

- Verified the split, drag-handle hover reveal, preview toggle, and mobile
  collapse/restore behavior across desktop (1280px) and mobile (375px)
  viewports.
- Snapshot baselines that include the editor layout will need regeneration
  (`bun run playwright:update`) since the divider changes the rendered output.


mohan-bee (approved):
veryy smoothh !! i love this

mohan-bee:
take review from others too

seveibar (commented):
it's pretty clear from the video there should be a minimum size for the pane! Otherwise users will be confused when they don't understand where the pane went

reviewing on mobile shortly

seveibar (approved):

PR #4321, opened 2026-08-09
Load Matchpack in the Solvers tab

mohan-bee:
### Motivation
- Make Matchpack solver debugging available on tscircuit.com.

### Before
- The site resolved an older Core registry without `LayoutPipelineSolver`.

### After
- The site resolves the current tscircuit, Core, and circuit-json versions with `LayoutPipelineSolver` in the shared registry.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,6 +4,10 @@
   "workspaces": {
     "": {
       "name": "@tscircuit/fake-snippets",
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


seveibar (changes requested):
Dont add deps!

mohan-bee:
> Dont add deps!

no i didn't revieed !! ai did that !!

PR #4552, opened 2026-08-22
Use shared Gerber file generation for fabrication downloads

seveibar:
## Summary
- consume convertCircuitJsonToGerberFiles from the dynamically loaded circuit-json-to-gerber package
- package the shared Gerber/Excellon file map with BOM and pick-and-place outputs
- include through-hole, blind/buried, and NPTH drill files
- add ZIP-content regression coverage

## Why
circuit-json-to-gerber@0.0.97 now owns complete Gerber and Excellon file generation. The site should package that authoritative file map rather than reproduce converter flags and drill-layer logic.

## Validation
- bun test bun-tests/download-fabrication-files.test.ts
- bun run typecheck
- Biome formatting
- git diff --check

### DIFF
--- a/bun-tests/download-fabrication-files.test.ts
+++ b/bun-tests/download-fabrication-files.test.ts
@@ -1,0 +1,48 @@
+import { expect, test } from "bun:test"
+import type { AnyCircuitElement } from "circuit-json"
+import JSZip from "jszip"
+import { createFabricationFilesZip } from "@/lib/download-fns/download-fabrication-files"
+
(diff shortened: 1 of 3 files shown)


seveibar:
Making this a dependent draft: the file-generation logic is being consolidated in tscircuit/circuit-json-to-gerber#149. Once that API is released, this PR should consume convertCircuitJsonToGerberFiles rather than maintain a site-specific Gerber/Excellon assembly path.

PR #4729, opened 2026-09-02
fix: export all schematic sheets in image downloads

hrithik18k:
## Motivation

Projects can contain multiple `schematic_sheet` elements, but schematic PNG and SVG downloads used `convertCircuitJsonToSchematicSvg`. That renderer intentionally selects one sheet, so multi-sheet projects downloaded only their first/default sheet—for example, the `host` sheet in `mohan-bee/hub`.

`circuit-to-svg` already provides `convertCircuitJsonToStackedSchematicSheetsSvg`, which renders every sheet in `sheet_index` order and preserves the existing behavior for projects with zero or one sheet.

## Changes

- use `convertCircuitJsonToStackedSchematicSheetsSvg` directly for schematic SVG downloads
- use the same stacked renderer before rasterizing schematic PNG downloads
- use the standard `image/svg+xml` MIME type for SVG downloads

## Result

Schematic PNG and SVG downloads now contain every sheet in one vertically stacked image. Single-sheet projects continue to render normally. PCB, assembly, pinout, and 3D downloads are unchanged.

## Testing

- `bun run typecheck`
- Biome formatting check on both changed files

### Before

https://github.com/user-attachments/assets/146e12e2-7694-46d5-9fd6-976c578cda0a

### After

https://github.com/user-attachments/assets/166b2962-147e-42c7-b056-ee5585cbfda4

### DIFF
--- a/src/lib/download-fns/download-circuit-png.ts
+++ b/src/lib/download-fns/download-circuit-png.ts
@@ -4,7 +4,7 @@ import {
   convertCircuitJsonToAssemblySvg,
   convertCircuitJsonToPcbSvg,
   convertCircuitJsonToPinoutSvg,
-  convertCircuitJsonToSchematicSvg,
+  convertCircuitJsonToStackedSchematicSheetsSvg,
 } from "circuit-to-svg"
 import { saveAs } from "file-saver"
 import { withDownloadToast } from "./download-toast"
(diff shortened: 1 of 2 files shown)


src/lib/download-fns/download-schematic-svg.test.ts
imrishabh18, line 1: This test doesn't make sense, why are you testing the method which is exported from the the library? Also, wrong place to add the test file

src/lib/download-fns/render-schematic-svg.ts
imrishabh18, line 1: No need for this new file, you can just replace the method in the main method directly

imrishabh18 (changes requested):

mohan-bee (approved):

hrithik18k:
Addressed both review comments. I removed the extra wrapper file and the test that only repeated circuit-to-svg coverage. The fix now directly replaces the renderer in the two existing SVG and PNG download paths. bun run typecheck and Biome formatting for both changed files pass. Thanks for the review!

imrishabh18 (approved):

PR #4762, opened 2026-09-04
update @tscircuit/footprinter to 0.0.426

mohan-bee:
### Motivation
- Let the 3D preview resolve supported footprint functions such as `smdpushbutton`.

### Before
- The site hoisted footprinter 0.0.102, which rejected newer footprint functions during 3D rendering.

### After
- The site resolves footprinter 0.0.426, and `smdpushbutton` produces four SMD pads.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,6 +4,9 @@
   "workspaces": {
     "": {
       "name": "@tscircuit/fake-snippets",
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


mohan-bee:
Closing because this added a new direct site dependency. The correction belongs in the package that already declares the outdated footprinter dependency.

PR #4874, opened 2026-09-10
fix: preserve package star state

KrishnaX12:
## Summary

- keep the star button state synced with the authenticated package query after SSR hydration
- support the existing `starred_by` package-list filter so profile pages can show starred packages
- add API and Playwright regression coverage for star, reload, unstar, and profile listing

## Root cause

Public SSR package data reports `is_starred: false`. The star hook copied that value into a separate query cache, which could remain stale after the authenticated package request completed. The profile page also sent `starred_by` to the fake snippets API, but that filter was not implemented.

## B4 : 

https://github.com/user-attachments/assets/2dc8b6a9-bafd-4b5f-9d36-8e71de97e68d

## After :

https://github.com/user-attachments/assets/04a03309-f8a7-46d5-aae0-9fbaaac7e335

## Testing

- `bun run typecheck`
- `bun test bun-tests/fake-snippets-api/routes/packages/list-1.test.ts bun-tests/fake-snippets-api/routes/packages/add_star.test.ts bun-tests/fake-snippets-api/routes/packages/remove_star.test.ts`
- `BUN_TEST=true bunx playwright test playwright-tests/star-ssr.spec.ts --workers=1`


imrishabh18 (commented):
rebase the PR, I merged a PR which makes sure that the fake is matching with the prod endpoint logic

imrishabh18 (approved):

